PR #7, opened 2016-05-18
Added client messages for map and cache partition iterator

eminn:


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java
```diff
@@ -719,4 +719,19 @@ Object addEntryListenerWithPredicate(String name, Data predicate, boolean includ
...
+     * The ordering of iteration over entries is undefined.
+     * java.util.Iterator#next() may return null if the entry is no longer present, has expired or has been evicted.
+     *
+     * @param name        Name of the map.
+     * @param partitionId The partition id which owns this record store.
+     * @param tableIndex  The slot number (or index) to start the iterator
+     * @param batch       The number of items to be batched
+     * @return last index processed and list of data
+     */
+    @Request(id = 60, retryable = false, response = ResponseMessageConst.MAP_KEY_ITERATOR_RESULT, partitionIdentifier = "partitionId")
```
asimarslan: I think it should be retryable

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
```diff
@@ -146,4 +146,12 @@ void Exception(int errorCode, String className, @Nullable String message, StackT
      */
     @Response(ResponseMessageConst.READ_RESULT_SET)
     void ReadResultSet(int readCount, List<Data> items);
+
+    /***
+     * @param tableIndex the last tableIndex processed
+     * @param keys       list of keys
+     */
+    @Response(ResponseMessageConst.MAP_KEY_ITERATOR_RESULT)
```
asimarslan: This one is identical to CacheKeyIteratorResult, why don't we just use it?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java
```diff
@@ -719,4 +719,19 @@ Object addEntryListenerWithPredicate(String name, Data predicate, boolean includ
 
     @Request(id = 59, retryable = false, response = ResponseMessageConst.VOID)
     void clearNearCache(String name, Address target);
+
+    /**
+     * Provides functionality to iterate over entries of a partition.
+     * The ordering of iteration over entries is undefined.
+     * java.util.Iterator#next() may return null if the entry is no longer present, has expired or has been evicted.
```
ihsandemir: Can you explain java.util.Iterator#next() without any java dependency? In the protocol document, we try not to be java specific. You can explain in terms of MapKeyIteratorResult since it shall go into the document as a new type. But what about "next" method, there is no such method for the protocol but the list is only converted to be an array of Data in the protocol document. So, may be we do not use next method name at all, you can say the next item in the array.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
```diff
@@ -146,4 +146,12 @@ void Exception(int errorCode, String className, @Nullable String message, StackT
      */
     @Response(ResponseMessageConst.READ_RESULT_SET)
     void ReadResultSet(int readCount, List<Data> items);
+
+    /***
+     * @param tableIndex the last tableIndex processed
```
ihsandemir: Will the user use this tableIndex in the next call to the iterate method? If so, let's put this info.

ihsandemir:
Need to generate the test binary resources e.g. https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/resources/1.0.protocol.compatibility.binary since it shall be version protocol 1.1 and put make sure that they are integrated into the tests for compatibility.

eminn:
@ihsandemir @asimarslan PR is updated, Could you guys have a final look at it ?

ihsandemir:
@eminn I think you also need to update the ResponseTemplate.java ResponseMessageConst.ENTRIES_WITH_CURSOR  to define the new response type that you introduced.

ihsandemir:
👍

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java
```diff
@@ -719,4 +718,29 @@ Object addEntryListenerWithPredicate(String name, Data predicate, boolean includ
...
+    Object fetchKeys(String name, int partitionId, int tableIndex, int batch);
+
+    /**
+     * Fetches specified number of entries from the specified partition starting from specified table index.
+     *
+     * @param name        Name of the map.
+     * @param partitionId The partition id which owns this record store.
+     * @param tableIndex  The slot number (or index) to start the iterator
+     * @param batch       The number of items to be batched
+     * @return last index processed and list of data
```
asimarslan: "List of entries" I guess

asimarslan:
👍

PR #11, opened 2016-06-15
Add since tag for the message codecs

ihsandemir:
Adds the since tag to message codecs so that we can mark since which version the request exists. It is used in protocol documentation.

### DIFF
--- a/hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodecCodeGenerator.java
+++ b/hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodecCodeGenerator.java
@@ -52,6 +52,7 @@ import javax.tools.StandardLocation;
 import java.io.IOException;
 import java.io.Serializable;
 import java.io.StringWriter;
+import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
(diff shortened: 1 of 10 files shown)


hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodecCodeGenerator.java
```diff
@@ -395,11 +440,15 @@ private void saveClass(String packageName, String className, String content) {
...
             JavaFileManager.Location location = StandardLocation.locationFor(StandardLocation.SOURCE_OUTPUT.name());
-            file = filer.createResource(location, packageName, fileName);
-            file.openWriter().append(content).close();
+            ProtocolFilePath path = new ProtocolFilePath(location, packageName, fileName);
+            Writer writer = openedFiles.get(path);
+            if (null == writer) {
+                writer = filer.createResource(location, packageName, fileName).openWriter();
+                openedFiles.put(path, writer);
+            }
+            writer.append(content);
```
mustafaiman: Where do we close theses files?

ihsandemir: @mustafaiman I actually put a writer close code at the end of process method, but this time  when it tries to access the writer to append again, it gives IOException that the writer is closed: 
java.io.IOException: Stream closed
        at sun.nio.cs.StreamEncoder.ensureOpen(StreamEncoder.java:45)
        at sun.nio.cs.StreamEncoder.write(StreamEncoder.java:118)
        at sun.nio.cs.StreamEncoder.write(StreamEncoder.java:135)
        at java.io.OutputStreamWriter.write(OutputStreamWriter.java:220)
        at java.io.FilterWriter.write(FilterWriter.java:91)
        at java.io.Writer.write(Writer.java:157)
        at java.io.Writer.append(Writer.java:227)
        at com.hazelcast.client.protocol.generator.CodecCodeGenerator.saveFile(CodecCodeGenerator.java:458)

I think that there is a possibility that the process may be called multiple time (Asim also "Processing code generator. round: " log which is probably due to this) Hence, I would not find a place where to put the file close code without any error. On finishing the run, I am assuming that the files are implicitly closed since I see the generated file has the content. 

I finally put it into finalize() method, but it does not execute this code at all, since it does not print the message that I put there:
        protected void finalize()
            throws Throwable {
        messager.printMessage(Diagnostic.Kind.WARNING, "Inside the finalize method");
        for (Writer writer : openedFiles.values()) {
            writer.close();
        }
    }

Hence, I decided to leave it so that it is implicitly closed. Let me know if any better suggestions on where we can put the close code.

mustafaiman: No better suggestion.

mustafaiman:
👍

PR #16, opened 2016-06-22
Add referenceId for the LockCodec

ihsandemir:
Added referenceId for the LockCodec. Updated the released jar version to 1.2.0.

Also adds <fieldname>Exist flag for any field that is an extension to the existing method parameters, e.g. see referenceId in LockCodecTemplate (LockLockCodec.java has a new field named referenceIdExist). The value of this field is set to false if the message is not encoded for protocol version 1.2. This is achieved with the clientMessage.isComplete() control before starting the first field of version 1.2 extensions. Hence, you can use referenceIdExist flag to check if the field exists or not in the decoded request when writing the message tasks. By the way, you may just use the java default value for the field if not exist, e.g. it is 0 for long type if the type does not exist, hence if you want the default to be 0, then you may not need to check for the referenceIdExist flag.


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/LockCodecTemplate.java
```diff
@@ -68,27 +68,30 @@
      * @param name      Name of the Lock
      * @param leaseTime Time to wait before releasing to lock
      * @param threadId  The id of the user thread performing the operation. It is used to guarantee that only the lock holder thread (if a lock exists on the entry) can perform the requested operation.
+     * @param referenceId The client-wide unique id for this request. It is used to make the request idempotent during retries.
```
asimarslan: Can you add a sentence on how we provide idempotency here. retries should be send with same referenceId etc.

asimarslan:
What abou map lock feature, do we need update for it?

ihsandemir:
@asimarslan Good point. I rechecked the PR https://github.com/hazelcast/hazelcast/pull/5602 . I see some changes at the map/impl/tx/TxnDeleteOperation and multimap/impl/txn/TxnCommitOperation. These probably matches the TransactionalMap and TransactionalMultiMap api. I will try to integrate that part as well.

Yes, I also need to make the changes to IMap and MultiMap lock api. I will go through the other structures including lock api.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/LockCodecTemplate.java
ihsandemir: Ok, changed the explanation like this:
"It is used to make the request idempotent by sending the same reference id during retries."

ihsandemir:
verify

hasancelik:
verify

hasancelik:
verify

hasancelik:
verify

hasancelik:
verify

hasancelik:
verify

hasancelik:
verify

hasancelik:
verify

hasancelik:
verify

ihsandemir:
verify

mustafaiman:
👍

PR #17, opened 2016-06-27
Change protocol version to digits at the document header page

ihsandemir:
Changed the protocol version written at the first page from 3 digits to 2 digits.


ihsandemir:
verify

hasancelik:
verify

hasancelik:
verify

PR #24, opened 2016-10-06
Need to exchange client and server Hazelcast versions for compatibility support implementation

ihsandemir:
Extended the authentication request with the clientHazelcastVersion and the response with the serverHazelcastVersion.


ihsandemir:
It is expected that this PR fails since it depends on the hazelcast master snaphot which does not have the needed changes yet. Due to the cyclic dependency, this PR will need to be merged before it becomes green, unless there is another better option to solve the cyclic dependency.

PR #35, opened 2016-11-09
Client Map codec extended to support aggregations and projections.

tkountis:


asimarslan:
@tkountis Can you give some info about the PR content. Is it for 3.8? if yes then we may need to change the version of CP to 1.4 as 1.3 is going to be  shipped with 3.7.3

tkountis:
@asimarslan sure, this is 3.8 part of the Fast Aggregations PRD.
I'll modify the `@since` annotations accordingly.

FYI, client implementation can be found here
https://github.com/hazelcast/hazelcast/pull/9185/commits/ea07833373deab162f078c5e7bf1578e59c4ed1a

asimarslan:
@tkountis There is only one  issue to validate :
Are the parameter of type Data (predicate, aggregator, projection) are all custom classes and no way to split them into parameters? Predicate is one that we used else where but I don't know the other parameters. Can you validate that we can not convert these into separate parameters ? 
If they are just user defined custom classes then we can use Data.

tkountis:
@asimarslan not quite sure I get the question, but if by separate parameters you mean their own codecs, then yes, I validate that these inputs (predicate, aggregator, projection) are *all* user implementated classes, so they need to be `Data`.

Let me know if I misunderstood the question.

CC'ing @tombujok

PR #37, opened 2016-11-15
Add protocol support of Null elements in Collections

tkountis:
Fix hazelcast/hazelcast#9240)

### DIFF
--- a/hazelcast-code-generator/src/main/java/com/hazelcast/annotation/MaybeNullElements.java
+++ b/hazelcast-code-generator/src/main/java/com/hazelcast/annotation/MaybeNullElements.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to denote if a collection might contain null elements.
+ * If nullable element one byte (1 if null , 0 if not null) is send before the element.
+ */
+@Documented
+@Retention(RetentionPolicy.SOURCE)
+@Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE })
+public @interface MaybeNullElements {
+}
--- a/hazelcast-code-generator/src/main/resources/codec-template-java.ftl
+++ b/hazelcast-code-generator/src/main/resources/codec-template-java.ftl
@@ -26,7 +26,7 @@ public final class ${model.className} {
         public static int calculateDataSize(<#list model.requestParams as param><@methodParam type=param.type/> ${param.name}<#if param_has_next>, </#if></#list>) {
             int dataSize = ClientMessage.HEADER_SIZE;
 <#list model.requestParams as p>
-    <@sizeText varName=p.name type=p.type isNullable=p.nullable/>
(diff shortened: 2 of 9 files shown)


hazelcast-code-generator/src/main/java/com/hazelcast/annotation/MaybeNullElements.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to denote if a collection might contain null elements.
+ * If nullable element one byte (1 if null , 0 if not null) is send before the element.
+ */
+@Documented
+@Retention(RetentionPolicy.SOURCE)
+@Target({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE })
+public @interface MaybeNullElements {
```
ahmetmircik: Maybe a different name can be used like `ContainsNullable` as a continuation of exiting name `Nullable`

ahmetmircik:
can we have tests for this PR?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
```diff
@@ -77,6 +78,12 @@
     void ListData(List<Data> response);
 
     /**
+     * @param response The operation result as an array of serialized byte-array that might have null entries.
+     */
+    @Response(ResponseMessageConst.LIST_DATA_MAYBE_NULL_ELEMENTS)
```
tkountis: I believe its better not to modify existing LIST_DATA since we can not guarantee that we do not break existing contracts. In my opinion a new response type seemed a better fit to differentiate between the two.

hazelcast-code-generator/src/main/java/com/hazelcast/annotation/MaybeNullElements.java
tkountis: We could, yes, good suggestion.

ahmetmircik: Also what will happen if i have both `@Nullable @MaybeNullElements` for a param like this:
 `void myMethod(@Nullable @MaybeNullElements List<Data> keys)`

tkountis:
@ahmetmircik thanks for the feedback. Regarding tests, I don't see any way to test this in terms of unit-tests, and it seems that there is no similar thing in the whole module. Ultimately, this would be tested as part of the integration with Hazelcast, through individual client tests. Last but not least, the generated code is compiling which one more reassurance. Since I may be wrong, I 'll ask @ihsandemir and @asimarslan opinion on that, if they have any suggestions on how to add further testing.

hazelcast-code-generator/src/main/java/com/hazelcast/annotation/MaybeNullElements.java
tkountis: I believe I tried it and it worked, but I don't remember for sure, I 'll give it another go and document the result here. Thanks!

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
ahmetmircik: ok

hazelcast-code-generator/src/main/resources/codec-template-java.ftl
```diff
@@ -26,7 +26,7 @@ public final class ${model.className} {
         public static int calculateDataSize(<#list model.requestParams as param><@methodParam type=param.type/> ${param.name}<#if param_has_next>, </#if></#list>) {
             int dataSize = ClientMessage.HEADER_SIZE;
 <#list model.requestParams as p>
-    <@sizeText varName=p.name type=p.type isNullable=p.nullable/>
+    <@sizeText varName=p.name type=p.type isNullable=p.nullable maybeNullElements=p.mayContainNullElements/>
```
ihsandemir: lets use the names consistent, using both maybeNullElements and mayContainNullElements is confusing.

hazelcast-code-generator/src/main/resources/codec-template-md.ftl
```diff
@@ -240,7 +240,7 @@ ${util.getReturnDescription(cm.comment)}
 | Name| Type| Nullable|Available since|
 |-------|------------|----------|-----|
         <#list cm.responseParams as param>
-|${param.name}| ${convertTypeToDocumentType(param.type)}| <#if param.nullable >Yes<#else>No</#if>|${param.sinceVersion}|
+|${param.name}| ${convertTypeToDocumentType(param.type)}| <#if param.nullable >Yes<#else>No</#if><#if param.containsNullable> / Contains Nullable</#if>|${param.sinceVersion}|
```
asimarslan: I think there should be one more line below this same as this one.

PR #39, opened 2016-11-25
Added and extended codecs for eventually consistent near-cache

ahmetmircik:
- Formatted some classes
- Added UUIDCodec
- Enhanced download.sh for the branch names with slash (e.g. fix/3.8/mybranch) 
- Added event, map, cache codes

PR #41, opened 2016-11-30
Revert: Added and extended codecs for eventually consistent near-cache

jerrinot:
Reasoning: #39 has a dependency on https://github.com/hazelcast/hazelcast/pull/9338
which is still in the review process.


ahmetmircik:
looks good.

jerrinot:
I had to add a new commit to increase the JSON lib - so the PR is not a pure revert of #39

asimarslan:
release 1.4.0-1

PR #42, opened 2016-11-30
Double read for config template causes cpp code generation to not work

ihsandemir:
Removed the mistakenly duplicated template read code.


asimarslan:
release 1.4.0-2

PR #48, opened 2016-12-08
Feature/master/csharp fix

asimarslan:
* fix client type constant check from "CSHARP" to "CSP" [BACKPORT]
* rename "varName" to "var_name" to overcome the freemarker bug


asimarslan:
release 1.4.0-5

PR #51, opened 2016-12-13
Remove uneeded args from API

tkountis:


asimarslan:
@tkountis released as 1.4.0-7

PR #55, opened 2017-01-13
Added removeAll codec to make implementation more client protocol fri…

ahmetmircik:
…endly


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java
```diff
@@ -816,4 +816,9 @@ Object addEntryListenerWithPredicate(String name, Data predicate, boolean includ
     @Since("1.4")
     Object assignAndGetUuids();
 
+
+    @Request(id = 68, retryable = false, response = ResponseMessageConst.VOID)
+    @Since("1.4")
+    void removeAll(String name, Data predicate);
```
asimarslan: Documentation is missing

ahmetmircik: fixed

PR #56, opened 2017-01-16
Cleanup codecs for continuous query

vbekiaris:
Rename EnterpriseMap-prefixed codecs to ContinuousQuery

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,8 +52,8 @@
     </modules>
 
     <properties>
-        <hazelcast.git.repo>ahmetmircik</hazelcast.git.repo>
-        <hazelcast.git.branch>fix/3.8/makeRemoveAllClientProtocolFriendly</hazelcast.git.branch>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -52,8 +52,8 @@
     </modules>
 
     <properties>
-        <hazelcast.git.repo>hazelcast</hazelcast.git.repo>
-        <hazelcast.git.branch>master</hazelcast.git.branch>
+        <hazelcast.git.repo>vbekiaris</hazelcast.git.repo>
```
Donnerbart: I think this change is a leftover from a local test.

vbekiaris: It's necessary for the build to pass, otherwise the build will download the code from `master` that still references `EnterpriseMap*` classes and will not compile. I think it's standard process to do this in a PR and switch back to `hazelcast/master` as soon as `master` changes are merged.

PR #57, opened 2017-01-24
Added addNearCacheInvalidationListener method to handle eventually co…

ahmetmircik:
…nsistent near cache related events


ahmetmircik:
run-lab-run

PR #60, opened 2017-02-09
Move logic for ScheduledExecutorService#getAllScheduled to the member side

tkountis:
Modify `ScheduledExecutorCodecTemplate#getAllScheduledFutures` to ignore address. 
Before we were issuing a Client -> Member request **per member** to fetch all scheduled tasks for a given scheduled executor. With this patch we are doing a **single** Client -> Member request, and let the member do the logic of fetching everything. 

Logic became a bit more complicated and is preferred to be kept on the member side, rather than having the protocol to deal with it. More details can be found here: hazelcast/hazelcast#9827


sancar:
@tkountis I have included your commit in my pr. We can close this one without merging.
This commit will be merged in https://github.com/hazelcast/hazelcast-client-protocol/pull/61

asimarslan:
This PR is move to #61 , no need to merge this one

PR #66, opened 2017-04-06
Add projection and filtering support for map iterator

mmedenjak:
The added projection and filtering support reuses the query engine but
allows running the query on chunk within a single partition. This
allows iteration with predicates and projections. Currently iterating
the indexes is not supported, the predicate will be run on each entry.
Also, the projection does not support the Offloadable which means that
it is invoked on the partition thread.

Changed the type of the entry which is provided to the predicate from
CachedQueryEntry to LazyMapEntry to allow the identity projection -
previously it failed when being serialized since CachedQueryEntry is
not serializable.

Also added some more generics and javadocs.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java
@@ -843,4 +843,20 @@ public interface MapCodecTemplate {
     @Since("1.4")
     Object addNearCacheInvalidationListener(String name, int listenerFlags, boolean localOnly);
 
+
+    /**
+     * Fetches the specified number of entries from the specified partition starting from specified table index
+     * that match the predicate and applies the projection logic on them.
+     *
+     * @param name       Name of the map
+     * @param tableIndex The slot number (or index) to start the iterator
+     * @param batch      The number of items to be batched
+     * @param projection projection to transform the entries with
+     * @param predicate  predicate to filter the entries with
+     * @return last index processed and list of entries after applied to the projection
+     */
+    @Request(id = 70, retryable = true, response = ResponseMessageConst.QUERY_RESULT_SEGMENT, partitionIdentifier = "partitionId")
+    @Since("1.5")
+    Object fetchWithQuery(String name, int tableIndex, int batch, Data projection, Data predicate);
+
 }
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/constants/ResponseMessageConst.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/constants/ResponseMessageConst.java
@@ -54,6 +54,7 @@ public final class ResponseMessageConst {
     @Since("1.4") public static final int ALL_SCHEDULED_TASK_HANDLERS = 121;
     @Since("1.4") public static final int NEAR_CACHE_INVALIDATION_META_DATA = 122;
     @Since("1.4") public static final int LIST_ENTRY_PARTITION_UUID = 123;
+    @Since("1.5") public static final int QUERY_RESULT_SEGMENT = 124;
 
     private ResponseMessageConst() {
     }
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
@@ -197,4 +197,12 @@ public interface ResponseTemplate {
     @Since("1.4")
     @Response(ResponseMessageConst.LIST_ENTRY_PARTITION_UUID)
     void PartitionUuidList(List<Map.Entry<Integer, UUID>> partitionUuidList);
+
+    /**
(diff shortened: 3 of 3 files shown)


mmedenjak:
Depends on https://github.com/hazelcast/hazelcast-client-protocol/pull/67 to turn green.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java
```diff
@@ -843,4 +843,21 @@ Object addEntryListenerWithPredicate(String name, Data predicate, boolean includ
...
+     * @param partitionId The partition id which owns this record store
+     * @param tableIndex  The slot number (or index) to start the iterator
+     * @param batch       The number of items to be batched
+     * @param projection  projection to transform the entries with
+     * @param predicate   predicate to filter the entries with
+     * @return last index processed and list of entries after applied to the projection
+     */
+    @Request(id = 70, retryable = true, response = ResponseMessageConst.ENTRIES_WITH_CURSOR, partitionIdentifier = "partitionId")
+    @Since("1.5")
+    Object fetchWithQuery(String name, int partitionId, int tableIndex, int batch, Data projection, Data predicate);
```
sancar: partition id is already send on clientMessage header. There is no need to send it again in message body.

sancar: You are not using this field in https://github.com/hazelcast/hazelcast/pull/10246 , as far as I see.

mmedenjak: Ok, will remove. I wasn't aware that the operation was run on the partition defined in the header. Seems like other fetch methods have the same problem but I guess we can't change that now : 
https://github.com/hazelcast/hazelcast-client-protocol/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java#L739
https://github.com/hazelcast/hazelcast-client-protocol/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java#L752

sancar: Yes, I saw them too, nothing we can do right now. 
I noted them in following prd to clean them up before 4.0
https://hazelcast.atlassian.net/wiki/display/PM/Client+Protocol+Enhancement+and+Cleanup

mmedenjak: Removed the argument. Should I remove the annotation element as well : `partitionIdentifier = "partitionId"`?

sancar: It should stay.

mmedenjak:
Please don't merge, the signature might change due to some PR comments in hazelcast

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java
```diff
@@ -843,4 +843,20 @@ Object addEntryListenerWithPredicate(String name, Data predicate, boolean includ
...
+     * that match the predicate and applies the projection logic on them.
+     *
+     * @param name       Name of the map
+     * @param tableIndex The slot number (or index) to start the iterator
+     * @param batch      The number of items to be batched
+     * @param projection projection to transform the entries with
+     * @param predicate  predicate to filter the entries with
+     * @return last index processed and list of entries after applied to the projection
+     */
+    @Request(id = 70, retryable = true, response = ResponseMessageConst.DATA, partitionIdentifier = "partitionId")
```
ihsandemir: why do you use Data for return type? As understand from javadoc, you can define it in ResponseTemplate.java and assign an id.

mmedenjak: I could but I thought this was a general preference for returning various return types, so the `ResponseTemplate` doesn't grow uncontrollably with various different types for different operations. And I saw that some tasks use this serialize->return Data->deserialize technique. 
Or is adding a new `ResponseTemplate` better?

ihsandemir: I think adding is better, since this way, it can be extended later without breaking compatibility. We prefer not using Data in the protocol.

ihsandemir: As I see from your description, the Collection<Integer> is the common part . List<Entry<Data,Data>> and Aggregator are changing between the two responses. What if you mark them as nullable? Would it help? Aggregator itself can be Data and that is how it is currently for other codecs. Keeping everything as a big Data will open way to break compatibility very easily.

mmedenjak: Updated to return only the projected map entries as `Data`.

PR #67, opened 2017-04-06
Generate build properties on downloaded hazelcast code

mmedenjak:
The build step was compiling hazelcast code but failed to generate
the GeneratedBuildProperties which causes the build to fail.
Added plugin for generating class on the same step as in the hazelcast
project.

See : https://github.com/hazelcast/hazelcast/commit/5fbbe09ee90777063d3c82f72ff00eebacbf1c0b

### DIFF
--- a/hazelcast/pom.xml
+++ b/hazelcast/pom.xml
@@ -151,7 +151,32 @@
                 </executions>
             </plugin>
 
-
+            <plugin>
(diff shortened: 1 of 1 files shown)


vbekiaris:
👍

mmedenjak:
@sancar can we merge this one? I don't seem to have write access.

PR #74, opened 2017-05-10
a new message added for ClientUserCodeDeployment

sancar:


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ClientMessageTemplate.java
```diff
@@ -173,4 +176,7 @@ Object authenticationCustom(Data credentials, @Nullable String uuid, @Nullable S
 
     @Request(id = 15, retryable = true, response = ResponseMessageConst.VOID)
     void ping();
+
```
ihsandemir: Missing javadoc.

PR #77, opened 2017-05-24
Generate correct code for response messages with primitive array

mmedenjak:
The code generation was expecting object arrays and was defaulting
the item value to `null`. This fails if we want to have a primitive
array. Here we default the value of the array item to the java
default value.


tkountis:
Referencing the macro in use:

```
<#macro getDefaultValueForType type>
<#if type == "boolean" >
false
<#elseif util.isPrimitive(type) >
0
<#else>
null
</#if>
</#macro>
```

sancar:
This problem is fixed here
https://github.com/hazelcast/hazelcast-client-protocol/pull/74
We can close this one.

PR #87, opened 2017-06-01
Fragment RingbufferService and support predicates and projections

mmedenjak:
Adapt RingbufferService to hold ringbuffers for other services (e.g.
map). This means that it should implement
FragmentedMigrationAwareService and return partition namespace data
for other services when requested.

Also, the read result sets were returning only how many items were
read as well as the matching items. In case the ringbuffer is read
with a filter, not all items match and the result set may not contain
items with a contiguous set of sequences. The itemCount is the
difference between the sequence ID of the last and first returned
item and is a hack not to send all sequence IDs.

We will send the sequence IDs instead. This allows the user to store
the sequence ID and resume processing even if the processing crashes
mid-point in the result set.

The ringbuffer already has supports for IFilter but we also add support
for predicates and projections.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
@@ -160,11 +160,27 @@ public interface ResponseTemplate {
             , int causeErrorCode, @Nullable String causeClassName);
 
     /**
-     * @param readCount Number of items in the response.
+     * @param readCount The number of items read from the ringbuffer. This can be different from
+     *                  the size of the item array if a filter was applied when reading and some
+     *                  items were skipped and are not in the returned array. This can also be
+     *                  seen as the delta with which the user should increase his sequence
+     *                  counter after consuming all of the returned items in the result set
+     *                  to request the next set from the ringbuffer.
      * @param items     The array of serialized items.
+     * @param itemSeqs  sequence IDs of returned ringbuffer items. This array can be {@code null}
+     *                  if the cluster version is 3.8 or lower. If the cluster version is 3.9 or
+     *                  higher then the array with sequence IDs will be sent. The array size is
+     *                  equal to the size of the items array and the arrays have a one-to-one
+     *                  mapping: the index of the sequence ID in the sequence array is equal
+     *                  to the index of the item in the item array. These sequences can be used to
+     *                  provide the user information to request a subset of the returned set if
+     *                  the user has for some reason stopped processing working when processing
+     *                  this returned set. If the ringbuffer is read without a filter then these
+     *                  sequences are a contiguous range and the size of the arrays is equal to the
+     *                  readCount.
      */
     @Response(ResponseMessageConst.READ_RESULT_SET)
-    void ReadResultSet(int readCount, List<Data> items);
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
```diff
@@ -162,9 +162,10 @@ void Exception(int errorCode, String className, @Nullable String message, StackT
     /**
      * @param readCount Number of items in the response.
      * @param items     The array of serialized items.
+     * @param itemSeqs  sequence IDs of returned ringbuffer items
      */
     @Response(ResponseMessageConst.READ_RESULT_SET)
-    void ReadResultSet(int readCount, List<Data> items);
+    void ReadResultSet(int readCount, List<Data> items, long[] itemSeqs);
```
ihsandemir: need Since Tag

mmedenjak:
@ihsandemir they used to be one PR but it's easier to review in two parts (at least the hazelcast part). 
If it is better for you, we can keep it as one PR in hazelcast-client-protocol.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
mmedenjak: Added and updated javadoc for `readCount`

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
```diff
@@ -160,11 +160,17 @@ void Exception(int errorCode, String className, @Nullable String message, StackT
...
+     *                  items were skipped and are not in the returned array. This can also be
+     *                  seen as the difference between the sequence ID of the last and first
+     *                  item in the returned array and the delta with which the user should
+     *                  increase his sequence counter to request the next set from the ringbuffer.
      * @param items     The array of serialized items.
+     * @param itemSeqs  sequence IDs of returned ringbuffer items
      */
     @Response(ResponseMessageConst.READ_RESULT_SET)
-    void ReadResultSet(int readCount, List<Data> items);
+    void ReadResultSet(int readCount, List<Data> items, @Since("1.5") @Nullable long[] itemSeqs);
```
ihsandemir: The extension seems to be ringbuffer specific. Extending the existing message will affect all users of READ_RESULT_SET. I would prefer designing a new response message like this:
        @Response(ResponseMessageConst.RINGBUFFER_READ_RESULT_SET)
        void ReadResultSet(int readCount, List<Map.Entry<Long, Data>> items);

mmedenjak: This PR is a preparation for the event journal PR : https://github.com/hazelcast/hazelcast-client-protocol/pull/88. I can even merge the changes into one PR - the split was made to make reviews on hazelcast easier.
In the event journal PR the map and cache templates use this response message and they need the sequence IDs. Also, ringbuffer users will need the sequence IDs once we add filtering to the reliable topic subscribers. Without the sequences for each item, the relablie topic subscriber or event journal reader cannot replay items from a certain point - it only knows the sequence ID of the first and last item.
Should I describe this in the javadoc or is there some change that I should make to this response message?

PR #94, opened 2017-08-02
Support for dynamically adding event journal config from client side

vbekiaris:
Required for https://github.com/hazelcast/hazelcast/pull/11019

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/DynamicConfigTemplate.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/DynamicConfigTemplate.java
@@ -381,4 +381,20 @@ public interface DynamicConfigTemplate {
                         @Nullable WanReplicationRef wanReplicationRef,
                         @Nullable HotRestartConfig hotRestartConfig);
 
+    /**
+     * Adds a new event journal configuration to a running cluster.
+     * If an event journal configuration for the same map or cache name already exists, then
+     * the new configuration is ignored and the existing one is preserved.
+     *
+     * @param mapName           name of {@code IMap} to use as event source
+     * @param cacheName         name of {@code ICache} to use as event source
+     * @param enabled           {@code true} to enable this event journal configuration, otherwise {@code false}
+     * @param capacity          capacity of event journal
+     * @param timeToLiveSeconds time to live (in seconds). This is the time the event journal retains items before removing them
+     *                          from the journal.
+     */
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/DynamicConfigTemplate.java
```diff
@@ -381,4 +381,20 @@ void addCacheConfig(String name, @Nullable String keyType, @Nullable String valu
...
+     *
+     * @param mapName           name of {@code IMap} to use as event source
+     * @param cacheName         name of {@code ICache} to use as event source
+     * @param enabled           {@code true} to enable this event journal configuration, otherwise {@code false}
+     * @param capacity          capacity of event journal
+     * @param timeToLiveSeconds time to live (in seconds). This is the time the event journal retains items before removing them
+     *                          from the journal.
+     */
+    @Request(id = 17, retryable = false, response = ResponseMessageConst.VOID)
+    void addEventJournalConfig(@Nullable String mapName, @Nullable String cacheName, boolean enabled, int capacity,
```
ihsandemir: is it meaningful to have both the mapName and cachName to be null?

vbekiaris:
@mmedenjak is it intended that both `mapName` and `cacheName` can be not-null in an `EventJournalConfig` ? or does it only make sense to have exclusively one or the other field non-null?

vbekiaris:
@mmedenjak says:
> Hi, regarding the journal config, I found it complicating if I restricted it to either map or cache (xsd and java validation) and no issue with allowing both. You can define one  journal config that matches both maps and caches with a patter. E.g. "journaled*"

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/DynamicConfigTemplate.java
mmedenjak: It isn't meaningful to have none. The user is free to define at-least-one : mapName, cacheName or both. Can we enforce this validation somehow?

vbekiaris: Not sure if that's possible to somehow express in the client-protocol, however I added validations on [client side](https://github.com/hazelcast/hazelcast/pull/11019/files#diff-56ee08fe8237fc406d39992cc8d36732R344) as well as on [member side](https://github.com/hazelcast/hazelcast/pull/11019/files#diff-cbf772471e0644a76279e71bca85d689R47)

PR #101, opened 2017-09-13
add partition state version to both event and response when fetching …

sancar:
…partitions


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/EventResponseTemplate.java
```diff
@@ -205,10 +205,11 @@ void IMapBatchInvalidation(List<Data> keys, @Since("1.4") List<String> sourceUui
                                @Since("1.4") List<UUID> partitionUuids, @Since("1.4") List<Long> sequences);
 
     /**
-     * @param partitions mappings from member address to list of partition id 's. Each event contains complete list
-     *                   of partitions.
+     * @param partitions            mappings from member address to list of partition id 's. Each event contains complete list
+     *                              of partitions.
+     * @param partitionStateVersion strictly increasing version of the partitions table.
      */
     @Since("1.5")
```
ihsandemir: since should be for the added new parameter only not the whole method.

sancar: Yeah but whole event is added in 1.5 not only the last parameter.

ihsandemir:
👍

PR #103, opened 2017-10-31
adding void delete to multimap

vmarquez:
This is to address issue 2719

https://github.com/hazelcast/hazelcast/issues/2719

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MultiMapCodecTemplate.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MultiMapCodecTemplate.java
@@ -266,4 +266,15 @@ public interface MultiMapCodecTemplate {
      */
     @Request(id = 21, retryable = false, response = ResponseMessageConst.BOOLEAN, partitionIdentifier = "key")
     Object removeEntry(String name, Data key, Data value, long threadId);
+
+    /**
+     * Removes all the entries with the given key.
+     *
+     * @param name     Name of the MultiMap
+     * @param key      The key of the entry to remove
+     * @param threadId The id of the user thread performing the operation. It is used to guarantee that only the lock holder thread (if a lock exists on the entry) can perform the requested operation
+     */
+    @Request(id = 22, retryable = false, response = ResponseMessageConst.VOID, partitionIdentifier = "key")
+    @Since("1.6")
+    void delete(String name, Data key, long threadId);
 }


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MultiMapCodecTemplate.java
```diff
@@ -266,4 +266,14 @@
...
     Object removeEntry(String name, Data key, Data value, long threadId);
+
+    /**
+     * Removes all the entries with the given key. Returns void
+     *
+     * @param name     Name of the MultiMap
+     * @param key      The key of the entry to remove
+     * @param threadId The id of the user thread performing the operation. It is used to guarantee that only the lock holder thread (if a lock exists on the entry) can perform the requested operation
+     */
+    @Request(id = 22, retryable = false, response = ResponseMessageConst.VOID, partitionIdentifier = "key")
```
mmedenjak: You may add the `@Since("1.5")` annotation here. Or is it 1.6? @ihsandemir @sancar

mmedenjak:
Looks good to me. I don't have write access to this repo so I guess the client team will have to approve this. @ihsandemir @sancar

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MultiMapCodecTemplate.java
vmarquez: @mmedenjak done.  I think 1.5 is right, I looked at other commits and it seemed others added 1.5 even when the .pom had the version already at 1.5.  LMK if I should change it.

ihsandemir: since 3.9 is released, can you use 1.6?

sancar: @vmarquez  Our next release will be 1.6 so tag should be 1.6.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MultiMapCodecTemplate.java
```diff
@@ -266,4 +266,15 @@
      */
     @Request(id = 21, retryable = false, response = ResponseMessageConst.BOOLEAN, partitionIdentifier = "key")
     Object removeEntry(String name, Data key, Data value, long threadId);
+
+    /**
+     * Removes all the entries with the given key. Returns void
```
ihsandemir: You may rather say that it does not return a response to the user. The protocol includes all client languages and void may not be very meaningful. You can also just not put this sentence since the response = ResponseMessageConst.VOID actually means the same thing.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MultiMapCodecTemplate.java
vmarquez: @sancar ah, makes sense, I have changed to 1.6.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MultiMapCodecTemplate.java
vmarquez: @ihsandemir thanks for the comments, I removed the part where it says `Returns void`.

asimarslan:
Hi @vmarquez ,

I'm planning to merge this PR. 
Can you please fill the [Hazelcast Contributor Agreement](https://hazelcast.atlassian.net/wiki/spaces/COM/pages/6357071/Hazelcast+Contributor+Agreement) and send to the email address on the page.
Thanks

vmarquez:
@asimarslan Just wanted to let you know I have emailed the address with a signed contributor agreement. LMK if you need something else done.  Thanks! :)

PR #104, opened 2017-11-07
version-changed check added to nodejs codec

mustafaiman:
Node.js codec template decodeResponse method missed `versionChanged` check. Therefore new codecs break backward compatibility by trying to read more than message has actually have. It throws exception.


mustafaiman:
verify

hasancelik:
verify

PR #106, opened 2017-11-13
Add message for Flake ID generator

viliam-durina:


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/FlakeIdGeneratorCodecTemplate.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.hazelcast.annotation.GenerateCodec;
+import com.hazelcast.annotation.Request;
+import com.hazelcast.client.impl.protocol.constants.ResponseMessageConst;
+
+@GenerateCodec(id = TemplateConstants.FLAKE_ID_GENERATOR_TEMPLATE_ID, name = "FlakeIdGenerator", ns = "Hazelcast.Client.Protocol.Codec")
+public interface FlakeIdGeneratorCodecTemplate {
+
+    @Request(id = 1, retryable = true, response = ResponseMessageConst.DATA)
+    Object newIdBatch(String name, int batchSize);
```
sancar: @viliam-durina 
We marking the messages when it will cause no harm to retry messages, in case we don't know the message we send consumed on members or not. For example, If operation is readonly / idempotent, then it is ok to mark as retryable=true.
If this is retried there will be some batch skipped and never used. There seems to be no harm to user. Was that the intend ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/TemplateConstants.java
```diff
@@ -51,6 +51,7 @@
     public static final int CARDINALITY_ESTIMATOR_TEMPLATE_ID = 28;
     @Since("1.4") public static final int SCHEDULED_EXECUTOR_TEMPLATE_ID = 29;
     @Since("1.5") public static final int DYNAMIC_CONFIG_TEMPLATE_ID = 30;
+    public static final int FLAKE_ID_GENERATOR_TEMPLATE_ID = 31;
```
sancar: We need `@Since("1.6")` here ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/FlakeIdGeneratorCodecTemplate.java
viliam-durina: yes, one batch will be lost if the message is retried, but there's no much harm in it.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/TemplateConstants.java
viliam-durina: Fixed

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/FlakeIdGeneratorCodecTemplate.java
```diff
@@ -0,0 +1,28 @@
...
+package com.hazelcast.client.impl.protocol.template;
+
+import com.hazelcast.annotation.GenerateCodec;
+import com.hazelcast.annotation.Request;
+import com.hazelcast.client.impl.protocol.constants.ResponseMessageConst;
+
+@GenerateCodec(id = TemplateConstants.FLAKE_ID_GENERATOR_TEMPLATE_ID, name = "FlakeIdGenerator", ns = "Hazelcast.Client.Protocol.Codec")
+public interface FlakeIdGeneratorCodecTemplate {
+
+    @Request(id = 1, retryable = true, response = ResponseMessageConst.DATA)
```
sancar: Response needs to be defined explicitly in protocol. Data should be used only for user data(key,value .. etc).  
As an example you can take look at:
https://github.com/hazelcast/hazelcast-client-protocol/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/MapCodecTemplate.java#L816
https://github.com/hazelcast/hazelcast-client-protocol/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java#L216

PR #113, opened 2018-02-05
Update PNCounter template

mmedenjak:
Adds replica timestamps (vector clock) to request and response
operations and adds a codec for adding dynamic config.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/PNCounterGeneratorCodecTemplate.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/PNCounterGeneratorCodecTemplate.java
@@ -20,33 +20,68 @@ import com.hazelcast.annotation.GenerateCodec;
 import com.hazelcast.annotation.Request;
 import com.hazelcast.annotation.Since;
 import com.hazelcast.client.impl.protocol.constants.ResponseMessageConst;
+import com.hazelcast.nio.Address;
+
+import java.util.List;
+import java.util.Map.Entry;
 
 @GenerateCodec(id = TemplateConstants.PN_COUNTER_ID_GENERATOR_TEMPLATE_ID, name = "PNCounter", ns = "Hazelcast.Client.Protocol.Codec")
 public interface PNCounterGeneratorCodecTemplate {
 
     /**
      * Query operation to retrieve the current value of the PNCounter.
+     * <p>
+     * The invocation will return the replica timestamps (vector clock) which
+     * can then be sent with the next invocation to keep session consistency
+     * guarantees.
+     * The target replica is determined by the {@code targetReplica} parameter.
+     * If smart routing is disabled, the actual member processing the client
+     * message may act as a proxy.
      *
-     * @param name the name of the PNCounter
+     * @param name              the name of the PNCounter
+     * @param replicaTimestamps last observed replica timestamps (vector clock)
+     * @param targetReplica     the target replica
      * @return the current value of the counter
      */
-    @Request(id = 1, retryable = true, response = ResponseMessageConst.LONG)
+    @Request(id = 1, retryable = true, response = ResponseMessageConst.CRDT_TIMESTAMPED_LONG)
     @Since("1.6")
-    Object get(String name);
+    Object get(String name, List<Entry<String, Long>> replicaTimestamps, Address targetReplica);
 
     /**
      * Adds a delta to the PNCounter value. The delta may be negative for a
      * subtraction.
+     * <p>
+     * The invocation will return the replica timestamps (vector clock) which
+     * can then be sent with the next invocation to keep session consistency
+     * guarantees.
+     * The target replica is determined by the {@code targetReplica} parameter.
+     * If smart routing is disabled, the actual member processing the client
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/PNCounterGeneratorCodecTemplate.java
```diff
@@ -17,36 +17,72 @@
...
+     * @param name              the name of the PNCounter
+     * @param replicaTimestamps last observed replica timestamps (vector clock)
+     * @param targetReplica     the target replica
      * @return the current value of the counter
      */
-    @Request(id = 1, retryable = true, response = ResponseMessageConst.LONG)
+    @Request(id = 1, retryable = true, response = ResponseMessageConst.CRDT_TIMESTAMPED_LONG)
     @Since("1.6")
-    Object get(String name);
+    Object get(String name, @Nullable List<Entry<String, Long>> replicaTimestamps, Address targetReplica);
```
sancar: What is the behavior when a null is provided for `replicaTimestamps`?
If it is not different than an empty list, I think it is better to remove Nullable. I have just checked the https://github.com/hazelcast/hazelcast/pull/11866, it may even simplify the implementation if we start with an empty VectorClock rather than a `null` one in `ClientPNCounterProxy`. What do you think?
If we expect a different behavior than an empty list, then the situation is different of course. And it needs to be explained in Javadoc for protocol developers also.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/PNCounterGeneratorCodecTemplate.java
```diff
@@ -17,36 +17,72 @@
...
+     *                          if it should return the value after the addition
+     * @param replicaTimestamps last observed replica timestamps (vector clock)
+     * @param targetReplica     the target replica
      * @return the value before the update if {@code getBeforeUpdate} is
      * {@code true}, otherwise the value after the update
      */
-    @Request(id = 2, retryable = false, response = ResponseMessageConst.LONG)
+    @Request(id = 2, retryable = false, response = ResponseMessageConst.CRDT_TIMESTAMPED_LONG)
+    @Since("1.6")
+    Object add(String name, long delta, boolean getBeforeUpdate, @Nullable List<Entry<String, Long>> replicaTimestamps, Address targetReplica);
```
sancar: @mmedenjak 
Similar comments for this Nullable also
see https://github.com/hazelcast/hazelcast-client-protocol/pull/113/files#r165902493

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/PNCounterGeneratorCodecTemplate.java
mmedenjak: No, there's no difference. I just wanted to transfer as little data as possible in case the proxy is just created and it hasn't yet observed any timestamps or secondly (and we don't have this yet), the client simply does not care about session consistency and continuously sends `null` to avoid the session consistency checks. An empty list would work as well.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/PNCounterGeneratorCodecTemplate.java
mmedenjak: There's no difference, used `null` to transfer as little data as possible.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/PNCounterGeneratorCodecTemplate.java
sancar: Difference between these is 3 bytes. 
If empty, 
In case of nullable, type information + a boolean denoting it is null. 
otherwise, type information + an integer denoting it is empty.  

If not empty, in nullable case, you will always have plus a boolean for every operation after some timestamps observed. 

So, If the data size is your concern, it depends on when timestamps observed. In the long run, nullable seems to be costing more. 

My concern was not the data size but simplicity(understandability) of the protocol. I would like to avoid letting a user do the same thing two different way. Also, the code in proxy seems to be simplified (if checks for the null case can be removed, I guess).

mmedenjak: Sure, will change.

mmedenjak: Updated.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/PNCounterGeneratorCodecTemplate.java
mmedenjak: Updated.

mmedenjak:
@ihsandemir can you also take a look since you have reviewed the original PN counter PR? Or can we merge this?

PR #115, opened 2018-02-09
Adds merge policy config for dynamic configs

vbekiaris:
Updates `DynamicConfigTemplate` to include `MergePolicyConfig`'s added to data structure configs in 3.10.


hazelcast-code-generator/src/main/resources/codec-template-md.ftl
```diff
@@ -495,6 +495,8 @@ Header only event message, no message body exist.
             <#return "array of string">
         <#case "java.util.List<java.lang.Long>">
             <#return "array of longs">
+        <#case "java.util.List<java.util.Map.Entry<java.lang.String,java.lang.Long>>">
+            <#return "array of key-value string,long pair">
```
ihsandemir: Is this only specific for key-value type string? What format is this string (like comma separated, how are they converted to string)?

vbekiaris: The type `List<Entry<String, Long>>` is used in `PNCounterGeneratorCodecTemplate` — it was introduced in a previous PR and the `String` key of the entries is a `UUID`, however this only applies for that particular use of the `List<Entry<String, Long>>` type.

I thought that it was an omission to not include it in `codec-template-md.ftl`, that’s why I added it with this generic description, if it's wrong let me know and I 'll remove this.

vbekiaris: As discussed on slack, this change is unrelated to the purpose of this PR and there are more data types which lack a description (the generated markdown lists them as `Unknown Data Type`s). So I have reverted this change and we will address missing type descriptions in a separate issue.

vbekiaris: Created issue #118

PR #123, opened 2018-02-22
fixes <java.lang.Long> conversion to Long

mustafaiman:
When java.lang.Long is used within a generic type in codec template, it was not converted to Node.js Client's `Long` primitive successfully. This PR fixes that problem. I bumped into this while generating PNCounter codecs.

Also, it includes some refactoring.


mustafaiman:
Checkstyle complains with `/scratch/jenkins/workspace/Client-Protocol-pr-builder/hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerationUtils.java:411:5: Cyclomatic Complexity is 16 (max allowed is 12).`
I think the method is pretty clean. I do not see any value in this checkstyle error.

mustafaiman:
verify

PR #124, opened 2018-02-26
Add next sequence to read from in ResultSet response

mmedenjak:
Sometimes the reader of the ringbuffer is loss tolerant and we may
allow the reader to be slow and skip some items. To indicate this, we
send an additional sequence to indicate where the reader can continue
reading from number of lost items so that the reader may check if
items have been lost (overwritten) and act on that information.


kwart:
run-lab-run

kwart:
verify

mmedenjak:
@sancar @ihsandemir can you please review (since you were reviewers for the original event journal PR)?

mmedenjak:
@sancar @asimarslan rebased

PR #126, opened 2018-03-08
Introducing createProxies to send proxies in batch

sancar:
related to https://github.com/hazelcast/hazelcast/pull/12539

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ClientMessageTemplate.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ClientMessageTemplate.java
@@ -354,4 +354,20 @@ public interface ClientMessageTemplate {
     @Request(id = 18, retryable = false, response = ResponseMessageConst.VOID, event = {EventMessageConst.EVENT_PARTITIONS})
     @Since(value = "1.5")
     void addPartitionListener();
+
+    /**
+     * Proxies will be created on all cluster members.
+     * If the member is  a lite member, a replicated map will not be created.
+     * Any proxy creation failure is logged on the server side.
+     * Exceptions related to a proxy creation failure is not send to the client.
+     * A proxy creation failure does not cancel this operation, all proxies will be attempted to be created.
+     *
+     * @param proxies proxies that will be created
+     *                Each entry's key is distributed object name.
+     *                Each entry's value is service name.
+     *                For possible service names see createProxy message.
+     */
+    @Since(value = "1.6")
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ClientMessageTemplate.java
```diff
@@ -354,4 +354,16 @@ Object authenticationCustom(Data credentials, @Nullable String uuid, @Nullable S
...
+     * @param target  target address to create proxies on
+     * @param proxies proxies that will be created
+     *                If member is lite member replicated map will not be created and will be ignored silently.
+     *                Each entry's key is distributed object name.
+     *                Each entry's value is service name.
+     *                For possible service names see createProxy message.
+     */
+    @Since(value = "1.6")
+    @Request(id = 19, retryable = false, response = ResponseMessageConst.VOID)
+    void createProxies(List<Map.Entry<String, String>> proxies, Address target);
```
ihsandemir: how will it work if some of the proxy creations fail?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ClientMessageTemplate.java
```diff
@@ -354,4 +354,16 @@ Object authenticationCustom(Data credentials, @Nullable String uuid, @Nullable S
     @Request(id = 18, retryable = false, response = ResponseMessageConst.VOID, event = {EventMessageConst.EVENT_PARTITIONS})
     @Since(value = "1.5")
     void addPartitionListener();
+
+    /**
+     * @param target  target address to create proxies on
```
ihsandemir: is this needed for dummy client?

sancar: Yes, it is needed. It is mainly introduced for replicated map/lite member problem. 
The client is trying to send to proxies to a non-lite member even though it's owner is a lite member.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ClientMessageTemplate.java
sancar: The only reason to fail is replicated map trying to be created on a lite member. And it is silently ignored. Behaviour is caused by `PostJoinProxyOperation` since that operation is used. 
But if one fails, the ones that failed is just logged on the server side. But all will be attempted to be created.

PR #135, opened 2018-04-20
event data version support added. C# codec generator added.

asimarslan:
we need hazelcast side fix for event handler versions

### DIFF
--- a/hazelcast-code-generator/src/main/resources/codec-template-java.ftl
+++ b/hazelcast-code-generator/src/main/resources/codec-template-java.ftl
@@ -1,4 +1,4 @@
-package ${model.packageName};
+ package ${model.packageName};
 
 import com.hazelcast.client.impl.protocol.ClientMessage;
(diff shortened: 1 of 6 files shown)


hazelcast-code-generator/src/main/resources/codec-template-java.ftl
```diff
@@ -141,35 +141,45 @@ public final class ${model.className} {
...
+                    if (clientMessage.isComplete()) {
+                        handle${event.name?cap_first}EventV${previousVersion}(${paramCallList});
+                        return;
                     }
                 </#if>
-                    if (!messageFinished) {
-                        <@readVariable var_name=p.name type=p.type isNullable=p.nullable isEvent=true />
-                    }
+                <#if p_index gt 0 ><#assign paramCallList=paramCallList + ", "></#if>
+                <#assign paramCallList += p.name>
```
sancar: We need following right after this line 
<#assign previousVersion = p.sinceVersion?replace('.','') >

It is available in C# part, looks like missed when porting to here.

PR #143, opened 2018-06-07
Adds ICache.SetExpiryPolicy codec

mustafaiman:


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/CacheCodecTemplate.java
```diff
@@ -441,4 +441,19 @@
...
+     *
+     * @param name          name of the cache
+     * @param keys          The keys that are associated with the specified expiry policy.
+     * @param expiryPolicy  custom expiry policy for this operation
+     * @param completionId  user generated id which shall be received as a field of the cache event upon completion of
+     *                      the request in the cluster.
+     */
+    @Request(id = 35, retryable = false, response = ResponseMessageConst.VOID)
+    @Since("1.7")
+    void setExpiryPolicy(String name, List<Data> keys, Data expiryPolicy, int completionId);
```
asimarslan: is `expiryPolicy` parameter `@Nullable` or not?

mustafaiman: it is not nullable. Is there an annotation for that? I assumed with non-nullable is implied without any annotation.

asimarslan: OK, I just make sure this parameter is not nullable. There is no annotation for it.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/CacheCodecTemplate.java
```diff
@@ -441,4 +441,19 @@
...
+     * {@code expiryPolicy} takes precedence for these particular {@code keys} against any cache wide expiry policy.
+     * If some keys in {@code keys} do not exist or are already expired, this call has no effect for those.
+     *
+     * @param name          name of the cache
+     * @param keys          The keys that are associated with the specified expiry policy.
+     * @param expiryPolicy  custom expiry policy for this operation
+     * @param completionId  user generated id which shall be received as a field of the cache event upon completion of
+     *                      the request in the cluster.
+     */
+    @Request(id = 35, retryable = false, response = ResponseMessageConst.VOID)
```
asimarslan: Can it be retryable ?

mustafaiman: No, this operation adds an expiry policy. Since it is an update to the record, it re-calculates the ttl of the record implicitly. Retrying the operation would make TTL unpredictable.

PR #166, opened 2019-01-29
Adds mapconfig#metadatapolicy

mustafaiman:
Adds new field to MapConfig

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/DynamicConfigTemplate.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/DynamicConfigTemplate.java
@@ -338,6 +338,8 @@ public interface DynamicConfigTemplate {
      *                                              or {@code null}
      * @param partitioningStrategyImplementation    a serialized instance of a partitioning strategy
      * @param hotRestartConfig                      hot restart configuration
+     * @param metadataPolicy                        metadata policy configuration for the supported data types. Valid values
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/DynamicConfigTemplate.java
```diff
@@ -355,7 +357,8 @@ void addMapConfig(String name, int backupCount, int asyncBackupCount, int timeTo
                       @Nullable String partitioningStrategyClassName,
                       @Nullable Data partitioningStrategyImplementation,
                       @Nullable HotRestartConfig hotRestartConfig,
-                      @Since("1.6") int mergeBatchSize);
+                      @Since("1.6") int mergeBatchSize,
+                      @Since("1.8") String preprocessing);
```
ihsandemir: In the doc, it mentions as if it can have only two values, CREATION_TIME or OFF, which seems to be an enumeration. We can better represent enumerations with Integer type.

mustafaiman: I used InMemoryFormat and EvictionPolicy as example. They are also encoded as strings.

ihsandemir: I think such uses were overseen during reviews. We always prefer enumeration for enumeration when possible (e.g. auth response code, etc.)

PR #179, opened 2019-03-26
Fix clean and download behaviour

sancar:
Clean was searching for a specific name and causing leftovers when
branch has changed. It will be downloaded to tempDownloaded directory
and all directory will be removed.

Download was working twice when mvn install is called. It is fixed
with <phase>DISABLE_FORKED_LIFECYCLE_MSOURCES-13</phase>

### DIFF
--- a/hazelcast/pom.xml
+++ b/hazelcast/pom.xml
@@ -105,9 +105,6 @@
                             <goal>exec</goal>
                         </goals>
                         <configuration>
-                            <arguments>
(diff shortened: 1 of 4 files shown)


sancar:
Fix for download twice with                         
`<phase>DISABLE_FORKED_LIFECYCLE_MSOURCES-13</phase>` found here
http://blog.peterlynch.ca/2010/05/maven-how-to-prevent-generate-sources.html

PR #187, opened 2019-06-11
Replace usages of c.h.core.Member with c.h.cluster.Member

mdogan:
See https://github.com/hazelcast/hazelcast/pull/15046

Uses branch: https://github.com/hazelcast/hazelcast/pull/15166

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,8 +60,8 @@
 
     <properties>
         <!-- Hazelcast branch to download and compile with -->
-        <hazelcast.git.repo>hazelcast</hazelcast.git.repo>
-        <hazelcast.git.branch>master</hazelcast.git.branch>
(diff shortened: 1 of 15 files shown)


pom.xml
```diff
@@ -60,8 +60,8 @@
 
     <properties>
         <!-- Hazelcast branch to download and compile with -->
-        <hazelcast.git.repo>hazelcast</hazelcast.git.repo>
-        <hazelcast.git.branch>master</hazelcast.git.branch>
+        <hazelcast.git.repo>mdogan</hazelcast.git.repo>
```
pveentjer: Is this correct?

asimarslan: yes, current cp needs custom branches before the merge.

mdogan: That's the only way for now, since client protocol and Hazelcast have a cyclic dependency to each other.

PR #190, opened 2019-06-27
Move merkle tree and event journal config inside map and cache config

mmedenjak:
In 4.0 we move the EventJournalConfig and MerkleTreeConfig inside
MapConfig and CacheConfig. Up until 4.0 there were top-level
configurations because CacheConfig was a BinaryInterface.

Also, because they are part of MapConfig and CacheConfig, they will no
longer have the mapName and cacheName fields.

Depends on: 
https://github.com/hazelcast/hazelcast/pull/15185
https://github.com/hazelcast/hazelcast/pull/15226


mmedenjak:
@sancar @ihsandemir are you ok with this PR, anything that I would need to change?

asimarslan:
Backported to new protocol branch

PR #196, opened 2019-07-30
Remove partitionId from executor.SubmitToPartition

sancar:
Partition id's already available on client message.
Requests should not carry a second partition id as parameter


sancar:
run-lab-run

sancar:
verify

PR #205, opened 2019-08-07
Code generator usage documentation

asimarslan:


README.md
```diff
@@ -0,0 +1,103 @@
...
+    name: METHOD-NAME-2
+    ...
+```
+
+A method(aka Remote method call) is defined by a request-response pair.
+
+A basic method structure example:
+
+```yaml
+  - id: METHOD-ID-1 (1-255)
```
mdumandag: The old protocol README (current master) explains what each property means. I think we should do the same for the new protocol too. (like what `since`, `retryable`, `acquires-resource` ... means)

README.md
```diff
@@ -0,0 +1,103 @@
...
+             Documentation of the parameter 2
+    response:
+      params:
+        - name: response parameter 1
+          type: Data
+          nullable: true
+          since: 2.0
+          doc: |
+             the response parameter 1
+
```
mdumandag: a sample `events` can be put here

README.md
```diff
@@ -0,0 +1,103 @@
...
+methods:
+  - id: METHOD-ID-1 (1-255)
+    name: METHOD-NAME-1
+    ...
+  - id: METHOD-ID-2 (1-255)
+    name: METHOD-NAME-2
+    ...
+```
+
+A method(aka Remote method call) is defined by a request-response pair.
```
mdumandag: we should also mention about the optional `events`

README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+## Code Generator
+
+The new protocol generator generates the related lang codec into the configured folder. It does not depend on hazelcast repo.
+
+
+### Setup
+
+You need the have python3 configured on your `PATH`. After cloning the repository, install the python library dependencies ones:
```
mdumandag: typo: `ones` should be removed

README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```bash
```
mdumandag: on some users, pip3 (pip backed by python 3) may be named as just `pip` I think we should just mention that this generator requires python 3 and change this to `pip install ...`

README.md
```diff
@@ -0,0 +1,103 @@
...
+pip3 install -r requirements.txt
+```
+
+### Code Generation
+
+You can simple generate for your favorite language by calling,
+
+```bash
+
+./generator.py [--root-dir ROOT_DIRECTORY] [--lang LANGUAGE]
```
mdumandag: for the same reason about the `pip3`, we should change this to `python generator.py ...`

README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+
+where `ROOT_DIRECTORY` is the root folder. If left empty default is `./output/[LANGUAGE]`  `LANGUAGE` is one of `JAVA`, `CPP`,
```
mdumandag: I think we should list the full names of the languages along with their abbreviations. It is not clear what `CS` or `TS` corresponds to

README.md
asimarslan: I think we can use schema doc for that purpose and just refer to it.

PR #213, opened 2019-08-20
Client backup acking

sancar:
Related to https://github.com/hazelcast/hazelcast/pull/14918


protocol-definitions/Client.yaml
```diff
@@ -903,3 +903,29 @@ methods:
...
+    doc: |
+      Adds listener for backup acks
+    request:
+      retryable: false
+      acquiresResource: false
+      partitionIdentifier: -1
+    response:
+      params:
+        - name: response
+          type: String
```
asimarslan: this should be a uuid

PR #219, opened 2019-09-11
New index configuration API (#15265)

devozerov:
The client protocol part of https://github.com/hazelcast/hazelcast/pull/15537


protocol-definitions/custom/Custom.yaml
```diff
@@ -174,15 +174,20 @@ customTypes:
...
         since: 2.0
-  - name: MapIndexConfig
+  - name: IndexConfig
     since: 2.0
+    returnWithFactory: true
     params:
-      - name: attribute
+      - name: name
         type: String
+        nullable: true
```
asimarslan: is `name`  parameter nullable?

devozerov: Yes, the index name is optional. If not provided by user, it will be constructed automatically on the server-side.

PR #223, opened 2019-09-17
Add Metrics protocol definition

blazember:
Definition of MetricsReadMetricsCodec.

### DIFF
--- a/protocol-definitions/Metrics.yaml
+++ b/protocol-definitions/Metrics.yaml
@@ -1,0 +1,35 @@
+id: 40
+name: Metrics
+methods:
+  - id: 1
+    name: readMetrics
+    since: 2.0
+    doc: |
+      Reads the recorded metrics starting with the smallest sequence number
+      greater or equals to the sequence number set in fromSequence.
+    request:
+      retryable: false
+      acquiresResource: false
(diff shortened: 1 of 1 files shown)


protocol-definitions/Metrics.yaml
```diff
@@ -0,0 +1,35 @@
...
+        - name: fromSequence
+          type: long
+          nullable: false
+          since: 2.0
+          doc: |
+            The sequence the recorded metrics should be read starting with.
+    response:
+      params:
+        - name: response
+          type: Data
```
asimarslan: what is the type of this Data? Can we expose it from protocol point of view? if it's just a simple class, you can list the parameters instead of reponse param or create a custom type ?

blazember: It may send either an exception, or a `com.hazelcast.internal.metrics.managementcenter.ConcurrentArrayRingbuffer.RingbufferSlice`. The latter is added in https://github.com/hazelcast/hazelcast/pull/15560/files#diff-f3e47ace2dc339bf108b5b8c1fb5d15eR126.

blazember:
Superseded by https://github.com/hazelcast/hazelcast-client-protocol/pull/230

PR #228, opened 2019-09-20
C# protocol generator

Scooletz:
This PR introduces C# (dotnet) generator for codecs

### DIFF
--- a/cs/codec-template.cs.j2
+++ b/cs/codec-template.cs.j2
@@ -15,16 +15,16 @@
 {%- endmacro %}
 {% macro decode_var_sized(param) -%}
     {%- if is_var_sized_list(param.type) or is_var_sized_list_contains_nullable(param.type) -%}
-        ListMultiFrameCodec.Decode{% if is_var_sized_list_contains_nullable(param.type) %}ContainsNullable{% endif %}{% if param.nullable  %}Nullable{% endif %}(ref iterator, {{ item_type(lang_name, param.type) }}Codec.Decode)
+        ListMultiFrameCodec.Decode{% if is_var_sized_list_contains_nullable(param.type) %}ContainsNullable{% endif %}{% if param.nullable  %}Nullable{% endif %}(iterator, {{ item_type(lang_name, param.type) }}Codec.Decode)
     {%- elif is_var_sized_entry_list(param.type) -%}
-        EntryListCodec.Decode(ref iterator, {{ key_type(lang_name, param.type) }}Codec.Decode, {{ value_type(lang_name, param.type) }}Codec.Decode)
+        EntryListCodec.Decode(iterator, {{ key_type(lang_name, param.type) }}Codec.Decode, {{ value_type(lang_name, param.type) }}Codec.Decode)
     {%- elif is_var_sized_map(param.type) -%}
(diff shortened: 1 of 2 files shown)


cs/codec-template.cs.j2
```diff
@@ -0,0 +1,264 @@
...
+        MapCodec.Decode{% if param.nullable  %}Nullable{% endif %}(ref iterator, {{ key_type(lang_name, param.type) }}Codec.Decode, {{ value_type(lang_name, param.type) }}Codec.Decode)
+    {%- else -%}
+        {%- if param.nullable  -%}
+            CodecUtil.DecodeNullable(ref iterator, {{ lang_name(param.type) }}Codec.Decode)
+        {%- else -%}
+            {{ lang_name(param.type) }}Codec.Decode(ref iterator)
+        {%- endif -%}
+    {%- endif -%}
+{%- endmacro %}
+/*
```
asimarslan: isn't `//` better for copyright notice?

cs/codec-template.cs.j2
```diff
@@ -0,0 +1,264 @@
...
+{% endif %}
+
+{% if namespace %}
+namespace {{ namespace }}
+{% else %}
+namespace Hazelcast.Client.Protocol.Codec
+{% endif %}
+{
+    /*
+    * This file is auto-generated by the Hazelcast Client Protocol Code Generator.
```
asimarslan: java style comment?

cs/codec-template.cs.j2
```diff
@@ -0,0 +1,264 @@
...
+    */
+
+    /// <summary>
+    {% for line in method.doc.splitlines() %}
+    /// {{ line }}
+    {% endfor %}
+    ///</summary>
+    internal static class {{ service_name|capital }}{{ method.name|capital }}Codec 
+    {
+        public const int RequestMessageType = {{ '0x%06X'|format(method.request.id) }};
```
asimarslan:
```    
//hex: {{ '0x%06X'|format(method.request.id) }}
public const int RequestMessageType = {{ method.request.id }};
```
using the hex as a comment and keep the value as decimal 
from Java codec. This was what we discussed to be best for debug.
You may consider if it apply to .net vs etc debuging cases too.

cs/codec-template.cs.j2
Scooletz: Will fix

cs/codec-template.cs.j2
Scooletz: Will fix

cs/codec-template.cs.j2
```diff
@@ -0,0 +1,266 @@
...
+                {% endfor %}
+                    Handle{{ event.name|capital }}Event({% for param in event.params %}{{param.name}}{% if not loop.last %}, {% endif %}{% endfor %});
+                    return;
+                }
+            {% endfor %}
+                Logger.GetLogger(GetType()).Finest("Unknown message type received on event handler :" + messageType);
+            }
+        {% for event in method.events%}
+
+            public abstract void Handle{{ event.name|capital }}Event({% for param in event.params %}{{ lang_types_encode(param.type) }} {{param.name}}{% if not loop.last %}, {% endif %}{% endfor %});
```
asimarslan: old protocol generator was using delegates, instead of abstract methods. overriding these methods on client code was not clean wrt passing method as parameters.

PR #230, opened 2019-09-23
Add Metrics Service definition

asimarslan:
Add Metrics Service definition and Introduce new protocol type Map_Long_byteArray
required for https://github.com/hazelcast/hazelcast/pull/15560

### DIFF
--- a/protocol-definitions/Metrics.yaml
+++ b/protocol-definitions/Metrics.yaml
@@ -1,0 +1,40 @@
+id: 39
+name: Metrics
+methods:
+  - id: 1
+    name: readMetrics
+    since: 2.0
+    doc: |
+      Reads the recorded metrics starting with the smallest sequence number
+      greater or equals to the sequence number set in fromSequence.
+    request:
+      retryable: true
+      acquiresResource: false
+      partitionIdentifier: -1
+      params:
+        - name: uuid
+          type: UUID
+          nullable: false
+          since: 2.0
+          doc: |
+            The UUID of the member that is supposed to read the metrics from.
+        - name: fromSequence
+          type: long
+          nullable: false
+          since: 2.0
+          doc: |
+            The sequence the recorded metrics should be read starting with.
+    response:
+      params:
+        - name: elements
+          type: Map_Long_byteArray
+          nullable: false
+          since: 2.0
+          doc: |
+            The map of timestamp and compressed metrics data
+        - name: nextSequence
+          type: long
+          nullable: false
+          since: 2.0
+          doc: |
+            The sequence number that the next task should start with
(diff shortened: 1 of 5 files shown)


protocol-definitions/Metrics.yaml
```diff
@@ -0,0 +1,40 @@
...
+      acquiresResource: false
+      partitionIdentifier: name
+      params:
+        - name: uuid
+          type: String
+          nullable: false
+          since: 2.0
+          doc: |
+            The UUID of the member that is supposed to read the metrics from.
+        - name: fromSequence
```
ihsandemir: better naming suggestion: `startSequenceNumber`

protocol-definitions/Metrics.yaml
```diff
@@ -0,0 +1,40 @@
...
+name: Metrics
+methods:
+  - id: 1
+    name: readMetrics
+    since: 2.0
+    doc: |
+      Reads the recorded metrics starting with the smallest sequence number
+      greater or equals to the sequence number set in fromSequence.
+    request:
+      retryable: false
```
ihsandemir: the call seems to be idempotent since it is a read-only method, hence, can it be retryable?

protocol-definitions/Metrics.yaml
```diff
@@ -0,0 +1,40 @@
...
+      acquiresResource: false
+      partitionIdentifier: name
+      params:
+        - name: uuid
+          type: String
```
ihsandemir: better to use `Uuid` as the type.

protocol-definitions/Metrics.yaml
```diff
@@ -0,0 +1,40 @@
...
+        - name: fromSequence
+          type: long
+          nullable: false
+          since: 2.0
+          doc: |
+            The sequence the recorded metrics should be read starting with.
+    response:
+      params:
+        - name: elements
+          type: Map_Long_byteArray
```
ihsandemir: Should the metrics be extended or should we use serialized byte array as here? byte array is always dangerous as it is very easy to break compatibility.

asimarslan: The byte array is the compressed data of all metrics. So it's by design byte[]

protocol-definitions/Metrics.yaml
asimarslan: It is the sequence number that it will continue to read from.

protocol-definitions/Metrics.yaml
asimarslan: @blazember is this an idempotent call?

protocol-definitions/Metrics.yaml
asimarslan: oh yes :)

blazember: Agree.

protocol-definitions/Metrics.yaml
blazember: Yes, it can be retryable.

PR #231, opened 2019-09-24
Remove legacy AtomicLong definitions

mdogan:
Hazelcast PR: https://github.com/hazelcast/hazelcast/pull/15601


protocol-definitions/AtomicLong.yaml
```diff
@@ -39,103 +46,68 @@ methods:
...
     response:
       params:
         - name: response
           type: long
           nullable: false
           since: 2.0
           doc: |
-            The old value before the function application.
-  - id: 5
+            TODO DOC
```
ihsandemir: TODO?

mdogan: I didn't add that. I just renamed `CPAtomicLong` to `AtomicLong`. And I see there are 290 `TODO DOC`s in the protocol repo. 🤷‍♂

PR #234, opened 2019-09-27
add binary file generator

mdumandag:
With this pr binary compatibility tests and their binaries are generated using `binary_generator.py`

This pr should be merged after https://github.com/hazelcast/hazelcast-client-protocol/pull/235


binary/reference_objects.py
```diff
@@ -0,0 +1,67 @@
...
+        self.least_sig_bits = least_sig_bits
+
+
+UUID = UuidHolder(123456789, 987654321)
+
+BYTEARRAY = bytearray([BYTE])
+
+LONGARRAY = [LONG]
+
+STRING = "localhost"
```
ihsandemir: what about utf8 specific characters?

PR #238, opened 2019-10-02
ownerless client

sancar:


protocol-definitions/Client.yaml
```diff
@@ -105,14 +92,7 @@ methods:
...
           doc: |
-            Unique string identifying the connected client uniquely. This string is generated by the owner member server
-            on initial connection. When the client connects to a non-owner member it sets this field on the request.
-        - name: ownerUuid
-          type: UUID
-          nullable: true
-          since: 2.0
-          doc: |
-            Unique string identifying the server member uniquely.
+            Unique string identifying the connected client uniquely.
```
asimarslan: the type is not a string anymore.

protocol-definitions/Client.yaml
```diff
@@ -28,20 +28,7 @@ methods:
...
-          since: 2.0
-          doc: |
-            Unique string identifying the server member uniquely.
-        - name: isOwnerConnection
-          type: boolean
-          nullable: false
-          since: 2.0
-          doc: |
-            You must set this field to true while connecting to the owner member, otherwise set to false.
+            Unique string identifying the connected client uniquely.
```
asimarslan: the type is not a string anymore.

PR #237, opened 2019-10-02
Add Protocol Definitions for Management Center Operations

emre-aydin:
Also merge `metrics.yaml` into `MC.yaml` which contains all MC related client operations.

PR #240, opened 2019-10-03
allow enum types to be used in codecs

mdumandag:
this pr allows enum types to be used in the codecs. This PR does not require a change in the Hazelcast side since there is no behaviour change for the existing definitions. It is meant to be used for the #237


asimarslan:
We won't support enum's in the protocol level, they'll only be used by internal custom type codecs

PR #242, opened 2019-10-04
Client statistics codec change

blazember:
Client `statistics` codec is changed as follows:
- A `long` `timestamp` parameter is added
- The parameter `stats` is renamed to `clientAttributes` as it will be carrying only key-value
pairs that are not metrics. 
- Metrics `byte[]` blob parameter carrying the collected metrics is added.

See the doc of the `statistics` codec definition including the description of the compressor algorithm.

IMDG PR: https://github.com/hazelcast/hazelcast/pull/15682

### DIFF
--- a/protocol-definitions/Client.yaml
+++ b/protocol-definitions/Client.yaml
@@ -611,166 +611,187 @@ methods:
     name: statistics
     since: 2.0
     doc: |
-      The statistics is a String that is composed of key=value pairs separated by ',' . The following characters
-      ('=' '.' ',' '\') should be escaped in IMap and ICache names by the escape character ('\'). E.g. if the map name is
-      MyMap.First, it will be escaped as: MyMap\.First
+      The statistics is composed of three parameters.
 
-      The statistics key identify the category and name of the statistics. It is formatted as:
-      mainCategory.subCategory.statisticName
+      The first paramteter is the timestamp taken when the statistics collected.
 
-      An e.g. Operating system committedVirtualMemorySize path would be: os.committedVirtualMemorySize
+      The second parameter, the clientAttribute is a String that is composed of key=value pairs separated by ','. The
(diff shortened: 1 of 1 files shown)


protocol-definitions/Client.yaml
```diff
@@ -649,166 +649,33 @@ methods:
     name: statistics
     since: 2.0
     doc: |
-      The statistics is a String that is composed of key=value pairs separated by ',' . The following characters
```
asimarslan: We need to add the compression/decompression algorithm doc here.

blazember: Added the description of the compression algorithm.

blazember:
@asimarslan updated the PR and its OS counterpart in https://github.com/hazelcast/hazelcast/pull/15682. Could you please take a look?

sancar:
@blazember 
Even tough metrics structure is well defined and can be described via the protocol itself, it seems we avoided doing so make it single byte[]. I assume the reason is compression. It feels like compression should be concern of protocol itself not individual messages. But that will require more work. 

Note that since metrics are defined outside of the protocol, we will not be able to use the extend mechanism of the protocol for metrics to support backward compatibility. Any addition/removed from metrics should be carefully done to make sure that all client versions can talk with all members versions. 

If compression is not a must or can be delayed to be implemented in protocol level, It would benefit us more to describe metrics structure in the protocol instead of using `byte[]`

sancar:
@blazember Talked with Asim. It seems we are aware of the backward compatibility concerns and ok with it. I am giving ok to both prs.

blazember:
Thank you for your review @sancar  :+1: 

As an additional note on the compression algo's description, the same is being used in the `MC.readMetrics` codec too. If there is a better place for describing it, let me know.

PR #245, opened 2019-10-10
[Draft]Client Fast Start - Extend the authentication response to include membership and partitions list

ihsandemir:
Extend the authentication response to include membership and partitions list.


ihsandemir:
Closing.

PR #255, opened 2019-10-31
MC: Add protocol definitions for member management operations

emre-aydin:
Core PR: https://github.com/hazelcast/hazelcast/pull/15916

### DIFF
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -222,3 +222,96 @@ methods:
             8 - FREE_NATIVE_MEMORY_SIZE
             9 - FREE_NATIVE_MEMORY_PERCENTAGE
     response: {}
+  - id: 5
+    name: getMemberConfig
+    since: 2.0
+    doc: |
+      Gets the effective config of a member rendered as XML.
+    request:
+      retryable: true
+      acquiresResource: false
+      partitionIdentifier: -1
+      params: []
+    response:
+      params:
+        - name: configXml
+          type: String
+          nullable: false
+          since: 2.0
+          doc: |
+            Effective config of a member rendered as XML.
+  - id: 6
+    name: runGc
+    since: 2.0
+    doc: |
+      Requests garbage collection to be performed on the member's JVM.
+    request:
+      retryable: false
+      acquiresResource: false
+      partitionIdentifier: -1
+      params: []
+    response: {}
+  - id: 7
+    name: getThreadDump
+    since: 2.0
+    doc: |
+      Takes a thread dump of the member's JVM.
+    request:
+      retryable: false
+      acquiresResource: false
+      partitionIdentifier: -1
+      params:
+        - name: dumpDeadLocks
+          type: boolean
+          nullable: false
+          since: 2.0
+          doc: |
+            Whether only dead-locked threads or all threads should be dumped.
+    response:
+      params:
+        - name: threadDump
+          type: String
+          nullable: false
+          since: 2.0
+          doc: |
+            Thread dump of the member's JVM.
+  - id: 8
+    name: shutdownMember
+    since: 2.0
+    doc: |
+      Shuts down a member.
+    request:
+      retryable: false
+      acquiresResource: false
+      partitionIdentifier: -1
+      params: []
+    response: {}
+  - id: 9
+    name: promoteMember
+    since: 2.0
+    doc: |
+      Promotes a lite member to a data member.
+    request:
+      retryable: false
+      acquiresResource: false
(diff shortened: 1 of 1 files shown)


protocol-definitions/MC.yaml
```diff
@@ -222,3 +222,96 @@ methods:
             8 - FREE_NATIVE_MEMORY_SIZE
             9 - FREE_NATIVE_MEMORY_PERCENTAGE
     response: {}
+  - id: 5
+    name: getMemberConfig
+    since: 2.0
+    doc: |
+      Gets the effective config of a member rendered as XML.
```
asimarslan: What about YAML?

protocol-definitions/MC.yaml
```diff
@@ -222,3 +222,96 @@ methods:
...
+    doc: |
+      Shuts down a member.
+    request:
+      retryable: false
+      acquiresResource: false
+      partitionIdentifier: -1
+      params: []
+    response: {}
+  - id: 9
+    name: promoteMember
```
asimarslan: as we are promoting the lite member, the name should be whether `protomoteLiteMember` or `promoteToMember`

protocol-definitions/MC.yaml
emre-aydin: We only show the effective configuration. AFAIK it is only available as XML.

protocol-definitions/MC.yaml
emre-aydin: `promoteLiteMember` makes sense, will update accordingly.

protocol-definitions/MC.yaml
```diff
@@ -222,3 +222,96 @@ methods:
...
+  - id: 5
+    name: getMemberConfig
+    since: 2.0
+    doc: |
+      Gets the effective config of a member rendered as XML.
+    request:
+      retryable: true
+      acquiresResource: false
+      partitionIdentifier: -1
+      params: []
```
asimarslan: For an empty param list, you can remove the params definition altogether.
example: [Client.ping](https://github.com/hazelcast/hazelcast-client-protocol/blob/master/protocol-definitions/Client.yaml#L600)

protocol-definitions/MC.yaml
```diff
@@ -222,3 +222,96 @@ methods:
...
+    response:
+      params:
+        - name: threadDump
+          type: String
+          nullable: false
+          since: 2.0
+          doc: |
+            Thread dump of the member's JVM.
+  - id: 8
+    name: shutdownMember
```
asimarslan: How does this method decide which member to shutdown?
From a protocol perspective, it's not clear. A unisocket or smart client may have different behaviour.
A member uuid parameter or a better documentation can be added.

emre-aydin: `Shuts down the member it's called on.` -> is this clear enough as documentation? Will add the same to the other which have the same behavior.

asimarslan: Yes the definition is clear for me.
In order to shut down a member,  you'll need to open a connection to it. That drops any unisocket support. 
If that's the case, we may put a note at the top as this service only supports smart client.

protocol-definitions/MC.yaml
```diff
@@ -222,3 +222,91 @@ methods:
...
+  - id: 9
+    name: promoteLiteMember
+    since: 2.0
+    doc: |
+      Promotes a lite member to a data member.
+    request:
+      retryable: false
+      acquiresResource: false
+      partitionIdentifier: -1
+    response: {}
```
ihsandemir: is the result of the promotion notified by exception only if it failed?

protocol-definitions/MC.yaml
emre-aydin: We only use smart client. Will add the note to the service about unisocket client.

protocol-definitions/MC.yaml
emre-aydin: yes

protocol-definitions/MC.yaml
emre-aydin: Thinking again, I don't see a reason why we should drop support for unisocket clients. We might need them for whatever reason in the future. I will add the `member` parameter to the operations that need it (also to the ones I added in my previous PR).

emre-aydin: After a discussion with Asim, we have decided to only document the requirement of smart client and keep the protocol definitions as they are.

emre-aydin: I've updated the documentation.

PR #256, opened 2019-11-06
MC: Add protocol definitions for client bw list operations

puzpuzpuz:
Core PR: https://github.com/hazelcast/hazelcast/pull/15942

### DIFF
--- a/schema/custom-codec-schema.json
+++ b/schema/custom-codec-schema.json
@@ -62,6 +62,7 @@
             "RingbufferStoreConfigHolder",
             "TimedExpiryPolicyFactoryConfig",
             "DurationConfig",
+            "ClientBwListEntry",
             "EntryList_String_String",
             "EntryList_String_byteArray",
             "EntryList_Long_byteArray",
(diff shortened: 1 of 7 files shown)


schema/custom-codec-schema.json
```diff
@@ -62,6 +62,7 @@
             "RingbufferStoreConfigHolder",
             "TimedExpiryPolicyFactoryConfig",
             "DurationConfig",
+            "ClientBwListEntry",
```
ihsandemir: Why not define `EntryList_int_String` which will be more reusable for other codecs if needed rather than making it MC specific?

puzpuzpuz: I have considered this option. I decided to use a custom codec instead of introducing the `EntryList_int_String` one as there is a pretty good chance that we'll need to add more fields into `ClientBwListEntry` type in the future.

puzpuzpuz: @ihsandemir Does it make sense to you? Or you still think it's worth it to introduce `EntryList_int_String` custom type?

puzpuzpuz:
@sancar as the core PR is merged, could you merge this one on Monday?

PR #257, opened 2019-11-06
MC: Add protocol definition for getTimedMemberState operation

emre-aydin:
Core PR: https://github.com/hazelcast/hazelcast/pull/15949

### DIFF
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -327,4 +327,4 @@ methods:
           nullable: true
           since: 2.0
           doc: |
-            Last TimedMemberState of the member, serialized as JSON.
+            Latest TimedMemberState of the member, serialized as JSON.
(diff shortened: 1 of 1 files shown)


protocol-definitions/MC.yaml
```diff
@@ -311,3 +311,20 @@ methods:
...
+      Gets the latest TimedMemberState of the member it's called on.
+    request:
+      retryable: true
+      acquiresResource: false
+      partitionIdentifier: -1
+    response:
+      params:
+        - name: timedMemberStateJson
+          type: String
+          nullable: true
```
sancar: How can this be null ?

emre-aydin: It's null when the member first starts and hasn't created a `TimedMemberState` yet, or MC is disabled on the member.

PR #258, opened 2019-11-07
C# generator

asimarslan:


cs/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+    "IndexConfig": "Hazelcast.Config.IndexConfig",
+    "AttributeConfig": "NA",
+    "ListenerConfigHolder": "NA",
+    "CacheSimpleEntryListenerConfig": "NA",
+
+    "Map_String_String": "IDictionary<string, string>"
+}
+
+_cs_types_encode = {
+    "CacheEventData": "NA",
```
mdumandag: Some types like `CacheEventData`, `List_Xid` ... marked as `NA` in both `_cs_types_encode ` and `_cs_types_decode`. Maybe we can move them to `_cs_types_common` so the same things are not written twice.

cs/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+        cs_type = _cs_types_decode[key]
+    except KeyError:
+        cs_type = _cs_types_common[key]
+    if cs_type == "NA":
+        raise NotImplementedError("Missing type Mapping")
+    return cs_type
+
+
+def cs_escape_keyword(value):
+    keywords = ["base"]
```
mdumandag: We can move this dictionary of escaped keywords outside to eliminate dictionary creation on each call

cs/codec-template.cs.j2
```diff
@@ -0,0 +1,269 @@
...
+            {% endfor %}
+                Logger.GetLogger(typeof(EventHandler)).Finest("Unknown message type received on event handler :" + messageType);
+            }
+        {% for event in method.events%}
+            public delegate void Handle{{ event.name|capital }}Event({% for param in event.params %}{{ lang_types_encode(param.type) }} {{param.name}}{% if not loop.last %}, {% endif %}{% endfor %});
+        {% endfor %}
+        }
+    {% endif %}
+    }
+}
```
mdumandag: Will there be a codec hash test for c# client ?

mdumandag:
Also, after merging this, we should update the jenkins job so that it also tries to generate c# codecs.

cs/codec-template.cs.j2
asimarslan: I think we don't need.

cs/__init__.py
asimarslan: This keyword list is also not complete, we may find a better way to escape.

cs/__init__.py
asimarslan: The reason is still there for not being in common, they'll need an encode and decode mappings in future but they are just not implemented. So let's keep it as is.

cs/__init__.py
asimarslan: I updated the full reserved words and moved to up

mdumandag: We may also convert keyword list to set since it contains more than 100 items now

PR #259, opened 2019-11-08
MC: Add cluster management operations

emre-aydin:
Following new operations are added:
* getClusterMetadata
* shutdownCluster
* changeClusterVersion

Core PR: https://github.com/hazelcast/hazelcast/pull/15961

### DIFF
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -384,3 +384,76 @@ methods:
           doc: |
             Client filter list entries.
     response: {}
+  - id: 14
+    name: getClusterMetadata
+    since: 2.0
+    doc: |
+      Gets the current metadata of a cluster.
+    request:
+      retryable: true
+      acquiresResource: false
+      partitionIdentifier: -1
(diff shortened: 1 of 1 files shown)


protocol-definitions/MC.yaml
```diff
@@ -328,3 +328,58 @@ methods:
...
+      partitionIdentifier: -1
+      params:
+        - name: majorVersion
+          type: byte
+          nullable: false
+          since: 2.0
+          doc: |
+            Major component of the new cluster version.
+        - name: minorVersion
+          type: byte
```
ihsandemir: does it need to work with customer specific release umbers or beta releases, an e.g. 4.1-Visa ? Is byte enough?

emre-aydin: we are only interested in major and minor version, as this is the cluster version, not member version. so, yes, byte is enough.

emre-aydin:
Thanks for the review @sancar ! I've merged the core PR, can you please merge this one?

PR #267, opened 2019-11-26
check service id only when the default protocol definitions is used

mdumandag:
When a custom protocol definitions path is used (for Jet, for example), the ordering of service ids are not required to be sequential. For example, Jet uses 254 for service id. If we add another codec to the hazelcast (like https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/internal/longregister/client/codec/LongRegister.yaml) with the id of 253 and add another codec to the jet with id of 252 there is no way for generator to be aware of that. So, if a custom protocol definitions path is used, we should simply not do any check for sequential service ids.


generator.py
```diff
@@ -95,7 +95,8 @@
 
 protocol_defs = load_services(protocol_defs_path)
 protocol_defs = sorted(protocol_defs, key=lambda proto_def: proto_def['id'])
-if not validate_services(protocol_defs, schema_path):
+check_service_id = proto_path_arg is None
```
ihsandemir: I prefer not using an argument which is designed for another purpose to implicitly imply "DO not check id". Therefore, I suggest that you put an explicit boolean option to turn it off.

PR #268, opened 2019-11-26
MC: Add protocol definitions for MC events support

puzpuzpuz:
Core PR: https://github.com/hazelcast/hazelcast/pull/16107

Depends on #266 and #275


protocol-definitions/MC.yaml
```diff
@@ -725,4 +724,20 @@ methods:
...
-            check is ignored because of the configuration
\ No newline at end of file
+            check is ignored because of the configuration
+  - id: 24
+    name: pollMCEvents
+    since: 2.0
+    doc: |
+      Polls events available on member. Once read, events are removed from member's internal queue.
+    request:
+      retryable: true
```
sancar: If request is not idempotent, it should not be retryable.

puzpuzpuz: Thanks for noticing this misconfiguration!

Yes, it's not idempotent. Going to fix it.

PR #269, opened 2019-11-27
added protocol definition for TransactionalQueue addAll method

alex-dukhno:
For now is WIP as effort to add bulk operation to `TransactionalQueue`


protocol-definitions/TransactionalQueue.yaml
```diff
@@ -203,3 +203,51 @@ methods:
           since: 2.0
           doc: |
             The number of elements in this collection
+  - id: 6
+    name: addAll
+    since: 2.0
```
sancar, line 208: Should be Since 2.1

PR #276, opened 2019-12-10
MC: Add protocol definitions for CP subsystem mgmt operations

puzpuzpuz:
Core PR: https://github.com/hazelcast/hazelcast/pull/16226

### DIFF
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -729,7 +729,8 @@ methods:
     name: pollMCEvents
     since: 2.0
     doc: |
-      Polls events available on member. Once read, events are removed from member's internal queue.
+      Polls events available on member. Once read, events are removed from
+      member's internal queue.
     request:
       retryable: false
       partitionIdentifier: -1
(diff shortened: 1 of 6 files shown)


protocol-definitions/MC.yaml
```diff
@@ -741,3 +742,55 @@ methods:
...
+    since: 2.0
+    doc: |
+      Returns the current list of CP members.
+    request:
+      retryable: true
+      partitionIdentifier: -1
+    response:
+      params:
+        - name: cpMembers
+          type: EntryList_String_String
```
sancar: UUID has its own type in the protocol. We should not represent them as string.

puzpuzpuz: Same applies to `Address`. But that means that I need to add at least one more custom data type, e.g. `EntryList_UUID_Address`. Does it sound OK to you?

sancar: It is ok.It is better to have correct types.

puzpuzpuz: Thanks. Going to fix it a bit later and ask you for another review round.

puzpuzpuz: Done.

PR #283, opened 2019-12-20
Add missing docs

mdumandag:
Added missing docs to the protocol definitions. Also, removed unused definitions: `CacheAssignAndGetUuidsCodec`, `ClientRemoveAllListenersCodec`, `MapClearNearCacheCodec` and `MapAssignAndGetUuidsCodec`

### DIFF
--- a/protocol-definitions/Cache.yaml
+++ b/protocol-definitions/Cache.yaml
@@ -5,8 +5,8 @@ methods:
     name: addEntryListener
     since: 2.0
     doc: |
-      Adds an entry listener for this cache. The listener will be notified
-      for all cache created/removed/updated/expired events.
+      Adds an entry listener for this cache. For the types of events that the listener
+      will be notified for, see the documentation of the type field of the Cache event below.
     request:
       retryable: false
       partitionIdentifier: -1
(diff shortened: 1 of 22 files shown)


protocol-definitions/Cache.yaml
```diff
@@ -5,7 +5,8 @@ methods:
     name: addEntryListener
     since: 2.0
     doc: |
-      TODO DOC
+      Adds an entry listener for this cache. The listener will be notified
+      for all cache created/removed/updated/expired events.
```
ihsandemir: better not to list the events here but refer to the below events list, they are more complete there.

protocol-definitions/Cache.yaml
```diff
@@ -724,7 +738,7 @@ methods:
     name: removeInvalidationListener
     since: 2.0
     doc: |
-      TODO DOC
+      Removes the specified invalidation listener. Returns silently if there is no such listener added before.
```
ihsandemir: What does "Returns silently" mean? You may say "no change is done in the cluster and returns ..." Does it return true in that case or false?

protocol-definitions/Client.yaml
```diff
@@ -377,24 +370,24 @@ methods:
...
             doc: |
-              TODO DOC
-  - id: 8
+              Address of the node that dispatches the event
+  - id: 7
     name: removePartitionLostListener
     since: 2.0
     doc: |
-      TODO DOC
+      Removes the specified partition lost listener. Returns silently if there is no such listener added before.
```
ihsandemir: silently --> clarify it. See above similar comment.

PR #297, opened 2020-01-20
added protocol definition for TransactionalQueue removeAll method

alex-dukhno:


protocol-definitions/TransactionalQueue.yaml
```diff
@@ -203,3 +203,44 @@ methods:
           since: 2.0
           doc: |
             The number of elements in this collection
+  - id: 6
+    name: removeAll
+    since: 2.0
```
sancar, line 208: Version should be 2.1.  versions in the parameters should be 2.1 as well.

PR #304, opened 2020-01-24
Support bitmap indexes

taburet:
OS: https://github.com/hazelcast/hazelcast/pull/16555

### DIFF
--- a/protocol-definitions/custom/Custom.yaml
+++ b/protocol-definitions/custom/Custom.yaml
@@ -190,6 +190,22 @@ customTypes:
         type: List_String
         nullable: false
         since: 2.0
+      - name: bitmapIndexOptions
+        type: BitmapIndexOptions
+        nullable: false
+        since: 2.0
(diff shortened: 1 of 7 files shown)


protocol-definitions/custom/Custom.yaml
```diff
@@ -190,6 +190,22 @@ customTypes:
         type: List_String
         nullable: false
         since: 2.0
+      - name: bitmapIndexOptions
+        type: BitmapIndexOptions
+        nullable: false
```
devozerov: I am concerned with `nullable: false` here. Options are not a mandatory part of index configuration, and implementors of a client protocol are free to leave them unsupported. But here we force users to implement this additional class anyway.

taburet: fixed in https://github.com/hazelcast/hazelcast-client-protocol/pull/304/commits/59b0fecec2f87f57dd11d9d257803c577c6bd143

PR #309, opened 2020-03-03
Add codec versioning support

mdumandag:
This PR adds the ability to add new parameters, methods, custom types and events to the client protocol. 

As discussed within the team, existence (or lack of it) of the new parameters can be checked by the `isXXExists` fields/arguments of the RequestParameters, ResponseParameters, handleYYEvent methods, custom type constructor calls/factory method calls.

Also, for the custom types, the following change is made. If the custom type does not have an initial frame before (no fix sized paramters), and we extend the custom type with new fix sized parameters, those parameters are encoded into the `BEGIN_FRAME` of the custom type since adding a new initial frame would break the old readers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ You can generate codecs for a specific language by calling,
 
 ```bash
 
-./generator.py [-r ROOT_DIRECTORY] [-l LANGUAGE] [-p PROTOCOL_DEFS_PATH] [-o OUTPUT_DIRECTORY] [-n NAMESPACE] [-b BINARY_OUTPUT_DIR] [-t TEST_OUTPUT_DIR] [-v VERSION] [--no-binary] [--no-id-check]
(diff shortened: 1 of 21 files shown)


schema/protocol-schema.json
```diff
@@ -6,7 +6,8 @@
   "definitions": {
     "since": {
       "enum": [
-        2.0
+        2.0,
+        2.1
```
mdumandag: Since we will have changes in the 2.0.1 branch, should we add it here ?

README.md
```diff
@@ -220,4 +219,29 @@ enum type to int is required. In the factory method, you will receive an integer
...
+Client protocol can be expanded by adding new
+* services
+* methods
+* parameters to existing method requests, responses or events
+* events to existing methods
+* custom types
+* parameters to existing custom types
+
+While expanding the protocol, one needs to follow these simple guidelines:
+* `since` field of the protocol definitions of the the newly added parameters, methods, events and custom types should
```
mdumandag: Where should we put the current protocol version ? Should it be on README, generator.py or somewhere else ?

asimarslan: The latest since tag will define the current version. Maybe a log during code generation will be helpful.

asimarslan: I don't think we should hard code else where.

PR #311, opened 2020-03-10
C# codec fix type parameter fixed

asimarslan:


zpqrtbnk:
@asimarslan Does this PR still make sense or should we kill it?

zpqrtbnk:
Killing.

PR #312, opened 2020-03-22
Add triggerMapLoader param for putAll

pertsodian:
See https://github.com/hazelcast/hazelcast/issues/5337
* triggerMapLoader = true: putAll
* triggerMapLoader = false: setAll

### DIFF
--- a/protocol-definitions/Map.yaml
+++ b/protocol-definitions/Map.yaml
@@ -1796,6 +1796,12 @@ methods:
           since: 2.0
           doc: |
             mappings to be stored in this map
+        - name: triggerMapLoader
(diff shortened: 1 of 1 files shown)


protocol-definitions/Map.yaml
```diff
@@ -1796,6 +1796,12 @@ methods:
           since: 2.0
           doc: |
             mappings to be stored in this map
+        - name: triggerMapLoader
+          type: boolean
+          nullable: false
+          since: 2.0
```
asimarslan, line 1802: 2.0 is released. This should be 2.1

PR #313, opened 2020-03-22
Add MultiMap.putAll interface (#9079)

abdulazizali77:
@mmedenjak @Holmistr 
please take a look

__Thoughts__

- This is generally copied from `IMap.putAll`. Have unit tests for client, and they 'work', but am unsure if this is completely correct, specifically related to partition mapping, and the response we should be returning.

### DIFF
--- a/binary/util.py
+++ b/binary/util.py
@@ -231,6 +231,7 @@ class VarSizedParamEncoder:
             'List_Integer': partial(FixSizedParamEncoder.encode_fix_sized_list_frame, item_type='int'),
             'List_Long': partial(FixSizedParamEncoder.encode_fix_sized_list_frame, item_type='long'),
             'List_UUID': partial(FixSizedParamEncoder.encode_fix_sized_list_frame, item_type='UUID'),
+            'List_Data': partial(self.encode_multi_frame_list, encoder=self.encode_data_frame),
             'List_ScheduledTaskHandler': partial(self.encode_multi_frame_list, encoder=self.encoder.custom_type_encoder
                                                  .encoder_for('ScheduledTaskHandler'))
         }
@@ -390,6 +391,7 @@ reference_objects_dict = {
     'EntryList_UUID_UUID': 'aListOfUUIDToUUID',
     'EntryList_UUID_List_Integer': 'aListOfUUIDToListOfIntegers',
     'EntryList_Data_Data': 'aListOfDataToData',
+    'EntryList_Data_List_Data': 'aListOfDataToListOfData',
     'Map_String_String': 'aMapOfStringToString',
     'List_byteArray': 'aListOfByteArrays',
     'List_CacheEventData': 'aListOfCacheEventData',
--- a/binary/__init__.py
+++ b/binary/__init__.py
@@ -81,6 +81,7 @@ VarLengthEntryListTypes = [
     'EntryList_String_EntryList_Integer_Long',
     'EntryList_UUID_List_Integer',
     'EntryList_Data_Data',
+    'EntryList_Data_List_Data',
 ]
 
 VarLengthMapTypes = [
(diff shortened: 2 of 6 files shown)


sancar:
@abdulazizali77 
The message should not carry any Data that carries hazelcast class. 
So the usage of DataCollection is not correct for the client protocol. 

For the client protocol, the fields should be readable separately. 
Collection of Data is hidden inside the Data of the  `EntryList_Data_Data` . It should have been `EntryList_Data_List_Data` . 
You are gonna need to add this new type `EntryList_Data_List_Data` to protocol schemas and types. 
For this, you need to add:
1. `EntryList_Data_List_Data` to `_java_types_decode` , `_java_types_encode`
2. 'EntryList_Data_List_Data': 'aListOfDataToListOfData', to `reference_objects_dict`
3. 'List_Data': partial(VarSizedParamEncoder.encode_data_frame, item_type='Data'), to `VarSizedParamEncoder`
4.  And in your original PR don't forget to put following to ReferenceObjects class.
public static List<Map.Entry<Data, List<Data>>> aListOfDataToListOfData  = Collections.singletonList(new AbstractMap.SimpleEntry<>(aData, aListOfData));

If I am not missing anything, you should be able to generate the code after these changes. After that, you can change the ClientMapProxyImpl to not use `DataCollection` class.

abdulazizali77:
thanks @sancar ill address these points, ill ping you if i need further clarifications

abdulazizali77:
@sancar please take a look,
had some issues with the new data type, so am not completely sure whether my solution is completely correct

__Issues__
- The new EntryList_Data_List_Data type has a List value type which isnt accounted for in the `is_var_sized_entry_list` template block, so either
a) add new `c.h.c.i.p.codec.builtin.ListDataCodec` to pass into `EntryListCodec.encode` (what was implemented)
b) add new custom type `List_Data` in Custom.yaml that will pass into `EntryListCodec.encode`
c) modify templates `is_var_sized_entry_list` block and possibly modify `EntryListCodec` or `ListMultiFrameCodec` classes
- currently the `test_MultiMapPutAllCodec_encodeRequest` are failing, and am investigating it, would appreciate some pointers

binary/util.py
```diff
@@ -231,6 +231,7 @@ def __init__(self, encoder):
             'List_Integer': partial(FixSizedParamEncoder.encode_fix_sized_list_frame, item_type='int'),
             'List_Long': partial(FixSizedParamEncoder.encode_fix_sized_list_frame, item_type='long'),
             'List_UUID': partial(FixSizedParamEncoder.encode_fix_sized_list_frame, item_type='UUID'),
+            'List_Data': partial(VarSizedParamEncoder.encode_data_frame),
```
mdumandag, line 234: This should be `'List_Data': partial(self.encode_multi_frame_list, encoder=self.encode_data_frame),` . Right now it is encoding a data frame instead of the list of data frames. That is the cause of the failing test

mdumandag:
For the other question, there are two possible solutions, one is the way you did it. (writing the builtin ListDataCodec).

Other way is the modifying the codec template such that we generate something like the following line for the cases that involve another list or entrylist as the key or value.

`EntryListCodec.encode(clientMessage, entries, DataCodec::encode, (msg, collection) -> ListMultiFrameCodec.encode(msg, collection, DataCodec::encode));` I think the first way is better, what do you think ? @sancar @asimarslan

binary/util.py
abdulazizali77: thanks @mdumandag ! i had an inkling it was the binary encoding but not sure what

sancar:
I think current approach is ok for now. If we need more complex structures, we may refactor to generate  `ListMultiFrameCodec` in the future. 

@abdulazizali77 is there any problem left ? Are your tests are passing ?

sancar:
verify

abdulazizali77:
> I think current approach is ok for now. If we need more complex structures, we may refactor to generate `ListMultiFrameCodec` in the future.
thanks @sancar 

> @abdulazizali77 is there any problem left ? Are your tests are passing ?
when i ran it last week with the new changes, my failing tests passed, i will look at the failing tests on the PR in bit

Holmistr:
Tests are passing. @mdumandag Could we get your approval as well? :)

abdulazizali77:
@Holmistr , sorry forgot to squash

PR #316, opened 2020-04-21
Supporting MemberInfo#addressMap

erosb:
# Supporting MemberInfo#addressMap

This PR adds support for the `MemberInfo#addressMap` property. We need this change for fixing a bug in MC.

Added custom types:
 * `Map_EndpointQualifier_Address` (type of `MemberInfo#addressMap`)
 * `EndpointQualifier` (key of the map)
 * `ProtocolType` (enum, part of `EndpointQualifier`)

## Related PRs and issues
 * IMDG PR: https://github.com/hazelcast/hazelcast/pull/16910
 * original problem: https://github.com/hazelcast/management-center/issues/2971
 * PR comment explaining why do we need correct MemberInfo#addressMap values: https://github.com/hazelcast/management-center/pull/3037#issuecomment-611439434


protocol-definitions/custom/Custom.yaml
```diff
@@ -780,6 +780,34 @@ customTypes:
...
+      - name: type
+        type: ProtocolType
+        nullable: false
+        since: 2.1
+      - name: identifier
+        type: String
+        since: 2.1
+        nullable: true
+  - name: ProtocolType
+    returnWithFactory: true
```
ihsandemir, line 800: do we need this?

erosb: Yes, if we delete this line then the generated code will try to instantiate with the `new` operator, which won't compile, since `ProtocolType` is an enum. So we need to pass the two attributes and implement an enum lookup method in `CustomTypeFactory`. See [in this PR](https://github.com/hazelcast/hazelcast/pull/16910/files#diff-35eafc96056b27133382f0fb9da039a0R197)

protocol-definitions/custom/Custom.yaml
```diff
@@ -780,6 +780,34 @@ customTypes:
         type: MemberVersion
         nullable: false
         since: 2.0
+      - name: addressMap
+        type: Map_EndpointQualifier_Address
+        nullable: false
+        since: 2.1
+  - name: EndpointQualifier
+    returnWithFactory: true
```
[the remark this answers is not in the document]
erosb: Yes. `EndpointQualifier` is not strictly an enum, but it has a pool of default instances, and we want to use those, instead of always creating a new instance during deserialization. So as far as I understand we need the factory.

PR #318, opened 2020-05-01
#4233 adding separate definitions for publishAll, publishAllAsync and publishAsync

andrewoelfing:
Please see issue https://github.com/hazelcast/hazelcast/issues/4233 for further information.
Core pr: https://github.com/hazelcast/hazelcast/pull/16946

### DIFF
--- a/protocol-definitions/Topic.yaml
+++ b/protocol-definitions/Topic.yaml
@@ -106,7 +106,7 @@ methods:
             True if registration is removed, false otherwise
   - id: 4
     name: publishAll
-    since: 2.0
+    since: 2.1
     doc: |
       Publishes all messages to all subscribers of this topic
     request:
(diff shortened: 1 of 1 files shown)


protocol-definitions/Topic.yaml
```diff
@@ -103,3 +104,83 @@ methods:
...
+          doc: |
+            The message to publish to all subscribers of this topic
+    response:
+      params:
+        - name: response
+          type: long
+          nullable: false
+          since: 2.0
+          doc: |
+            the CompletionStage to synchronize on completion.
```
sancar, line 157: @andrewoelfing 
I believe this is the registration id, not the CompletionStage , right ?
Also check response of `publishAllAsync`

protocol-definitions/Topic.yaml
```diff
@@ -103,3 +104,83 @@ methods:
           since: 2.0
           doc: |
             True if registration is removed, false otherwise
+  - id: 4
+    name: publishAll
+    since: 2.0
```
sancar, line 109: New additions should have `since: 2.1` .

protocol-definitions/Topic.yaml
andrewoelfing: Hello @sancar, what do you mean with the registration id? There is no response when the event is published (please see `EventService#publishEvent`) . Maybe this can the number of currently published messages to the topic?

protocol-definitions/Topic.yaml
andrewoelfing: Ok. :)

protocol-definitions/Topic.yaml
sancar: `type: long` got me confused. There should not be any response. Can you make this part 
`response {}` as it is in publishAll/publish ? Same goes for `publishAllAsync` as well.

andrewoelfing: Yes, I will change it.

PR #322, opened 2020-05-19
Remove unused codes from codecs

mdumandag:
Our eslint config does not allows unused variables, so,
RESPONSE_PARAMS which is there for debugging purposes is converted
into a comment.

Also, since nullable UUIDs are handled inside the FixSizedTypesCodec,
there is no need to import CodecUtil in that case.

depends on #321

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ where
     * `cpp` : C++
     * `cs` : C#
     * `py` : Python
-    * `ts` : Typescript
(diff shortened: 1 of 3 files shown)


puzpuzpuz:
@mdumandag could you finalize and merge this PR once you're back from vacation? I've already merged the node.js client couterpart PR, so it would be nice to have the protocol in sync.

PR #324, opened 2020-05-20
ProtocolType enum is removed from protocol custom types

asimarslan:
ProtocolType enum is removed from protocol custom types and handled as builtin enum handling (as int)

### DIFF
--- a/binary/reference_objects.py
+++ b/binary/reference_objects.py
@@ -48,6 +48,9 @@ objects = {
     },
     'BitmapIndexOptions': {
         'uniqueKeyTransformation': 1
+    },
+    'EndpointQualifier': {
+        'type': 2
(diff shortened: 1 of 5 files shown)


binary/reference_objects.py
```diff
@@ -48,9 +48,6 @@ def __init__(self, most_sig_bits, least_sig_bits):
     },
     'BitmapIndexOptions': {
         'uniqueKeyTransformation': 1
-    },
-    'ProtocolType': {
```
mdumandag, line 52: We should add `    'EndpointQualifier': {
        'type': 2
    }` since right now the binary encoder is writing 25 (default value for int) to the binary files but there is no enum for the id of 25.

PR #326, opened 2020-06-02
Add CPSubsystem listener definitions

mdogan:
See https://hazelcast.atlassian.net/wiki/spaces/PM/pages/2201026733/CP+Subsystem+Membership+Listener

Hazelcast PR: https://github.com/hazelcast/hazelcast/pull/17037

### DIFF
--- a/protocol-definitions/CPSubsystem.yaml
+++ b/protocol-definitions/CPSubsystem.yaml
@@ -1,0 +1,133 @@
+id: 34
+name: CPSubsystem
+methods:
+  - id: 1
+    name: addMembershipListener
+    since: 2.1
+    doc: |
+      Registers a new CP membership listener.
+    request:
+      retryable: false
+      partitionIdentifier: -1
+      params:
+        - name: local
+          type: boolean
+          nullable: false
+          since: 2.1
+          doc: |
+            Denotes whether register a local listener or not.
+    response:
+      params:
+        - name: response
+          type: UUID
+          nullable: false
+          since: 2.1
+          doc: |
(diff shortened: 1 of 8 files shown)


protocol-definitions/CPSubsystem.yaml
```diff
@@ -0,0 +1,133 @@
...
+        since: 2.1
+        params:
+          - name: member
+            type: CPMember
+            nullable: false
+            since: 2.1
+            doc: |
+              Member which is added or removed.
+          - name: type
+            type: String
```
asimarslan, line 38: Client protocol uses int values for enum types. int value will be faster and occupy less space. For non-java languages, int values are more convenient than string as they are usually just cast to enum types(contrary to java).
see: https://github.com/hazelcast/hazelcast-client-protocol/blob/master/protocol-definitions/Map.yaml#L1211
for usage.
an int value is not the ordinal of the enum at this point, it just refers to a value mapping and with the doc.

mdogan: This is a very rare event, so a few bytes of space efficiency is not that important. Adding another int type to an enum is redefining it, then we have to maintain both enum's name and type field and also a method in `FixedSizeTypesCodec`. 

If enum to int mapping is a must, I think I'll just change the api and use int instead of enum.

mdogan:
@asimarslan & @sancar: If you don't have any objections, can we you please merge this?

sancar:
@mdogan We can merge as soon as the original core pr is merged

PR #328, opened 2020-07-02
Get rid of ResponseParameters for single response

sancar:
ClientCodec returns the parameters on a different object.
This is necessary for a multivariable response but for single
types it is an unnecessary object.

I haven’t touched the requests since it is not common to have
a single type request. And they are not critical ones.

Core pr: https://github.com/hazelcast/hazelcast/pull/17162

### DIFF
--- a/java/binary/client-binary-compatibility-template.j2
+++ b/java/binary/client-binary-compatibility-template.j2
@@ -85,6 +85,11 @@ public class ClientCompatibility{% if test_nullable %}Null{% endif %}Test_{{ '_'
     public void test_{{ service.name|capital }}{{ method.name|capital }}Codec_decodeResponse() {
         int fileClientMessageIndex = {{ counter.count }};
         ClientMessage fromFile = clientMessages.get(fileClientMessageIndex);
+        {% set singleResponseValue = method.response.params|length == 1 and response_new_params|length == 0 %}
+        {% if singleResponseValue %}
+        {% set param = method.response.params|last %}
+        assertTrue(isEqual({%if test_nullable and param.nullable %}null{% else %}{{ reference_objects_dict[param.type] }}{% endif %}, {{ service.name|capital }}{{ method.name|capital }}Codec.decodeResponse(fromFile)));
+        {% else %}
         {{ service.name|capital }}{{ method.name|capital }}Codec.ResponseParameters parameters = {{ service.name|capital }}{{ method.name|capital }}Codec.decodeResponse(fromFile);
         {% set new_response_params = new_params(method.since, method.response.params) %}
         {% for param in method.response.params %}
@@ -97,6 +102,7 @@ public class ClientCompatibility{% if test_nullable %}Null{% endif %}Test_{{ '_'
         assertTrue(isEqual({%if test_nullable and param.nullable %}null{% else %}{{ reference_objects_dict[param.type] }}{% endif %}, parameters.{{ param.name }}));
             {% endif %}
         {% endfor %}
+        {% endif %}
     }
     {% set counter.count = counter.count + 1 %}
     {% if method.events|length != 0%}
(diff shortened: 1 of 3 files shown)


java/binary/client-binary-compatibility-template.j2
```diff
@@ -170,3 +176,5 @@ public class ClientCompatibility{% if test_nullable %}Null{% endif %}Test_{{ '_'
         }
     }
 }
+
+
```
mdumandag, line 180: Minor: The last newline is unnecessary

java/codec-template.java.j2
```diff
@@ -361,3 +392,4 @@ public final class {{ service_name|capital }}{{ method.name|capital }}Codec {
 {% endif %}
 }
 
+
```
mdumandag, line 426: Minor: This newline is unncessary

java/codec-template.java.j2
```diff
@@ -198,26 +198,38 @@ public final class {{ service_name|capital }}{{ method.name|capital }}Codec {
...
+{% if singleResponseValue %}
+    {% set param = method.response.params|last %}
+    /**
+    {% for line in param.doc.splitlines() %}
+     * {{ line }}
+    {% endfor %}
+     */
+    @edu.umd.cs.findbugs.annotations.SuppressFBWarnings({"UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD"})
+    public {% if param.nullable  %}@Nullable {% endif %}{{ lang_types_decode(param.type) }} {{ param.name }};
+{% else %}
```
mdumandag, line 242: If the `method.response.params|length` is equal to 0, there is no need to define ResponseParameters class and `decodeResponse` method. Maybe as an enhancement, we could handle that too. (For example, MapSetCodec is such a codec). However, If we do that, we should also fix binary compatiibility test template along with binary file generator. (They should skip generating tests/binary content for the empty response cases)

sancar: I will remove the ones with no parameter also. I don't intend to change the binary file generator. I will leave placeholder methods for ones with zero parameters as follows:

```
    public void test_ClientCreateProxyCodec_decodeResponse() {
        int fileClientMessageIndex = 9;
    }
```

mdumandag: Okay, we could revisit that later

java/codec-template.java.j2
```diff
@@ -196,28 +229,43 @@ public final class {{ service_name|capital }}{{ method.name|capital }}Codec {
...
     @edu.umd.cs.findbugs.annotations.SuppressFBWarnings({"URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD"})
     public static class ResponseParameters {
-{% for param in method.response.params %}
+    {% for param in method.response.params %}
+            /**
```
mdumandag, line 251: Can you decrease the indent level by one for the javadoc and fields ? Right now, RequestParameters has 4 spaces for their fields whereas ResponseParams has  8.

sancar: I could not find where the problem is. Generated code also looks correct. 
Request and Response Parameters have 8 spaces if they are multivalue(because they inside another class). 
When they are single value both have 4 spaces.

sancar: Ok I saw the problem. 👍

java/codec-template.java.j2
```diff
@@ -196,28 +229,43 @@ public final class {{ service_name|capital }}{{ method.name|capital }}Codec {
...
          * {{ line }}
         {% endfor %}
          */
         public {% if param.nullable  %}@Nullable {% endif %}{{ lang_types_decode(param.type) }} {{ param.name }};
-{% endfor %}
-{% for param in response_new_params %}
+    {% endfor %}
+    {% for param in response_new_params %}
+
+            /**
```
mdumandag, line 260: This part also has the same spacing problem

PR #329, opened 2020-07-03
SQL client support (protocol) (#17173)

devozerov:
This PR introduces client-side support for SQL. The design is explained in the issue https://github.com/hazelcast/hazelcast/issues/17173#issue-650589250

Closes https://github.com/hazelcast/hazelcast/issues/17173


protocol-definitions/custom/Custom.yaml
```diff
@@ -868,5 +868,74 @@ customTypes:
           type: Data
           nullable: true
           since: 2.0
-
-
+  - name: SqlQueryId
+    since: 2.1
+    params:
+      - name: memberIdHigh
+        type: long
```
sancar, line 875: We prefer to use UUID type for the sake of being uniform. 
All other representations of UUID's directly use UUID type.

devozerov: `QueryId` is an internal object. It is constructed from several UUIDs, but it is not UUID itself. I think uniformity is not applicable here - the object is designed this way on the server-side for a reason.

devozerov: Resolving, since the query ID is a string now.

protocol-definitions/custom/Custom.yaml
```diff
@@ -868,5 +868,44 @@ customTypes:
...
+    params:
+      - name: name
+        type: String
+        nullable: false
+        since: 2.1
+      - name: type
+        type: int
+        nullable: false
+        since: 2.1
+  - name: SqlRowMetadata
```
asimarslan, line 898: if this is a single parameter custom type, you can unwrap it

protocol-definitions/custom/Custom.yaml
```diff
@@ -868,5 +868,44 @@ customTypes:
...
+    since: 2.1
+    params:
+      - name: columns
+        type: List_SqlColumnMetadata
+        nullable: false
+        since: 2.1
+  - name: SqlRow
```
asimarslan, line 905: if this is a single parameter custom type, you can unwrap it

protocol-definitions/custom/Custom.yaml
devozerov: Will do.

devozerov: SqlRowMetadata has been removed.

protocol-definitions/custom/Custom.yaml
devozerov: I tried to remove the `SqlRow` from the protocol. If I do so, I would have to introduce the new `List_List_Data` type, because currently there is `List_SqlRow` as well. 

The problem is that in the future we will not pass row content as `List_Data`. Instead, it will be some form of `SqlRow` with some different content (e.g. specially encoded `SqlRowValue` objects or so). When we do this change, `List_List_Data` will be unused. So IMO it is better to use `SqlRow` for now, because it will not lead to unused types in the future.

WDYT?

asimarslan: First of all, client protocol doesn't support changing the custom type. We can only add more parameters.
If you are planning to update a Data type with a full custom encoding then, you need to update method id.
List_List_Data is currently supported, all you need to do is add it to init list like others. The code generator will handle the generation and no new codec will be added.
it will generate:
        ListMultiFrameCodec.encode(clientMessage, listDataData, ListDataCodec::encode);
where ListMultiFrameCodec and ListDataCodec  are already in the codebase.
When you need to update List_List_Data to List_SqlRow you just delete List_List_Data if it's not used elsewhere.
please note that List_List_Data is some generic codec that CP generator handles smartly.
From pure CP best practices, it would be better to add an actual SqlRow with actual fields later on.

protocol-definitions/custom/Custom.yaml
```diff
@@ -868,5 +868,74 @@ customTypes:
...
+        nullable: false
+        since: 2.1
+  - name: SqlRowMetadata
+    since: 2.1
+    params:
+      - name: columns
+        type: List_SqlColumnMetadata
+        nullable: false
+        since: 2.1
+  - name: SqlPageRow
```
asimarslan: This one is also a single parameter custom type. Can you just use List_Data instead of `SqlPageRow` and remove this custom codec

protocol-definitions/custom/Custom.yaml
```diff
@@ -868,5 +868,74 @@ customTypes:
...
+    params:
+      - name: values
+        type: List_Data
+        nullable: false
+        since: 2.1
+  - name: SqlPage
+    since: 2.1
+    params:
+      - name: rows
+        type: List_SqlPageRow
```
asimarslan: When you remove List_SqlPageRow, this can be List_Data_Data

protocol-definitions/custom/Custom.yaml
devozerov: Perhaps you noticed this in some outdated version. The is no such type at the moment.

protocol-definitions/custom/Custom.yaml
```diff
@@ -868,5 +868,49 @@ customTypes:
...
+  - name: SqlQueryId
+    since: 2.1
+    params:
+      - name: memberIdHigh
+        type: long
+        nullable: false
+        since: 2.1
+      - name: memberIdLow
+        type: long
+        nullable: true
```
asimarslan, line 880: is this nullable?

devozerov: Fixed (copy-paste artifact). It seems that this flag doesn't affect the generated code anyhow, so perhaps it could be made optional for primitive types?

PR #343, opened 2020-09-15
Simplify SqlExecute response

viliam-durina:
Should be merged only if hazelcast/hazelcast#17550 is merged.


sancar (approved):
Ok assuming that we don't need to give compatibility with 4.1 BETA release.

PR #344, opened 2020-09-16
Remove response wrapping in case of single param in Node.js client

puzpuzpuz:
Corresponding Node.js client PR: https://github.com/hazelcast/hazelcast-nodejs-client/pull/599

Also includes the following:
* Simplifies noResponseValue in Node.js and Java templates
* Simplifies check in Node.js and Java templates

### DIFF
--- a/ts/codec-template.ts.j2
+++ b/ts/codec-template.ts.j2
@@ -110,7 +110,12 @@ const RESPONSE_{{ to_upper_snake_case(param.name) }}_OFFSET = {% if loop.first %
 const EVENT_{{ to_upper_snake_case(event.name)}}_{{to_upper_snake_case(param.name)}}_OFFSET = {% if loop.first %}PARTITION_ID_OFFSET + BitsUtil.INT_SIZE_IN_BYTES{% else %}EVENT_{{ to_upper_snake_case(event.name)}}_{{ to_upper_snake_case(loop.previtem.name)}}_OFFSET + BitsUtil.{{loop.previtem.type.upper()}}_SIZE_IN_BYTES{% endif %};
     {% endfor %}
 {% endfor %}
-{% if method.response.params|length > 0 %}
+{#RESPONSE PARAMETERS#}
+{% set noResponseValue = method.response.params|length == 0 and response_new_params|length == 0 %}
+{% set singleResponseValue = method.response.params|length == 1 and response_new_params|length == 0 %}
+{% if noResponseValue %}
+{% elif singleResponseValue %}
+{% else %}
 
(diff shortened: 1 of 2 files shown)


ts/codec-template.ts.j2
```diff
@@ -110,7 +110,12 @@ const RESPONSE_{{ to_upper_snake_case(param.name) }}_OFFSET = {% if loop.first %
 const EVENT_{{ to_upper_snake_case(event.name)}}_{{to_upper_snake_case(param.name)}}_OFFSET = {% if loop.first %}PARTITION_ID_OFFSET + BitsUtil.INT_SIZE_IN_BYTES{% else %}EVENT_{{ to_upper_snake_case(event.name)}}_{{ to_upper_snake_case(loop.previtem.name)}}_OFFSET + BitsUtil.{{loop.previtem.type.upper()}}_SIZE_IN_BYTES{% endif %};
     {% endfor %}
 {% endfor %}
-{% if method.response.params|length > 0 %}
+{#RESPONSE PARAMETERS#}
+{% set noResponseValue = method.response.params|length == 0 and response_new_params|length == 0 %}
```
mdumandag, line 114: Minor: `{% set noResponseValue = method.response.params|length == 0 %}` should be enough here since response_new_params should be of length 0 when there is no response params

ts/codec-template.ts.j2
```diff
@@ -110,7 +110,12 @@ const RESPONSE_{{ to_upper_snake_case(param.name) }}_OFFSET = {% if loop.first %
...
+{% set singleResponseValue = method.response.params|length == 1 and response_new_params|length == 0 %}
+{% if noResponseValue %}
+{% elif singleResponseValue %}
+{% else %}
```
mdumandag, line 118: Minor: Can we replace this part with something like `{% if not (noResponseValue or singleResponseValue) %} `?

mdumandag (approved):
Approving with minor comments

ts/codec-template.ts.j2
puzpuzpuz: TBH I was looking at `codec-template.java.j2` when I was doing this change. But it makes sense. Simplified templates for both Node.js and Java in 09e9223a5fc039a34185823880674bc17f2b72a5

ts/codec-template.ts.j2
puzpuzpuz: Simplified for both Node.js and Java templates in 658ca896698d705354c8af0434cd5b878b908c88

mdumandag (approved):

PR #353, opened 2020-10-12
Improve ignorelist functionality

sancar:
ignore_service_list is with names instead of id's
Adds ability to ignore on operation level.
if one adds "Map.put" to list, it will ignore one
operation.


cpp/__init__.py
```diff
@@ -1,5 +1,6 @@
+cpp_ignore_service_list = {"Cache", "XATransaction", "ContinuousQuery", "DurableExecutor", "CardinalityEstimator",
```
ihsandemir, line 1: CPSubsystem should not be ignored for C++.

ihsandemir (approved):

mdumandag (approved):

PR #356, opened 2020-10-28
Add Migration Listener

peterjot:
Depends on https://github.com/hazelcast/hazelcast/pull/17713
Issue https://github.com/hazelcast/hazelcast/issues/16260

### DIFF
--- a/protocol-definitions/Client.yaml
+++ b/protocol-definitions/Client.yaml
@@ -775,3 +775,111 @@ methods:
       partitionIdentifier: -1
     response:
       {}
+  - id: 17
+    name: addMigrationListener
+    since: 2.1
+    doc: |
+      Adds a Migration listener to the cluster.
+    request:
+      retryable: false
+      partitionIdentifier: -1
+      params:
+        - name: localOnly
+          type: boolean
+          nullable: false
+          since: 2.1
+          doc: |
+            if true only node that has the migration sends the request, if false
+            sends all partition lost events.
+    response:
+      params:
+        - name: response
+          type: UUID
+          nullable: false
+          since: 2.1
+          doc: |
+            The listener registration id.
+    events:
+      - name: Migration
+        since: 2.1
+        params:
+          - name: migrationState
+            type: MigrationState
+            nullable: false
+            since: 2.1
+            doc: |
+              Migration state.
+          - name: partitionId
+            type: int
+            nullable: false
+            since: 2.1
+            doc: |
+              The id assigned during the listener registration.
+      - name: ReplicaMigration
+        since: 2.1
+        params:
+          - name: migrationState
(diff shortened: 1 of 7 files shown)


protocol-definitions/Client.yaml
```diff
@@ -774,4 +774,112 @@ methods:
...
+      - name: Migration
+        since: 2.1
+        params:
+          - name: migrationState
+            type: MigrationState
+            nullable: false
+            since: 2.1
+            doc: |
+              Migration state.
+          - name: partitionId
```
sancar, line 812: `partition id` name is confusing. Can you name it as `event type`? 
Can you also list possible values on the doc part?  We generally use positive values. 
MIGRATION_STARTED 0
MIGRATION_COMLETED 1

protocol-definitions/Client.yaml
```diff
@@ -774,4 +774,112 @@ methods:
...
+      params:
+        - name: response
+          type: UUID
+          nullable: false
+          since: 2.1
+          doc: |
+            The listener registration id.
+    events:
+      - name: Migration
+        since: 2.1
```
sancar, line 804: The version should be `2.2`. Can you replace all your additions accordingly?

protocol-definitions/Client.yaml
```diff
@@ -774,4 +774,112 @@ methods:
...
+    request:
+      retryable: false
+      partitionIdentifier: -1
+      params:
+        - name: localOnly
+          type: boolean
+          nullable: false
+          since: 2.2
+          doc: |
+            if true only node that has the migration sends the request, if false
```
puzpuzpuz, line 792: nit: consider changing to:
```
If set to true, the server adds the listener only to itself, otherwise the listener is is added for all members in the cluster.
```

protocol-definitions/Client.yaml
```diff
@@ -774,4 +774,112 @@ methods:
...
+          doc: |
+            The id assigned during the listener registration.
+    response:
+      params:
+        - name: response
+          type: boolean
+          nullable: false
+          since: 2.2
+          doc: |
+            true if the listener existed and removed, false otherwise.
```
puzpuzpuz, line 885:
```suggestion
            true if the listener existed and was removed, false otherwise.
```

puzpuzpuz (approved):

protocol-definitions/Client.yaml
peterjot: Sound better, thanks  👍

peterjot:
Thanks for the code review guys, I believe this PR is close to being finalized!

puzpuzpuz:
@sancar I forgot to merge this one. Thanks for merging it!

PR #360, opened 2020-12-09
Schema support for the SQL public API

devozerov:
The client protocol part of the https://github.com/hazelcast/hazelcast/pull/17953


ihsandemir (approved):

devozerov:
@ihsandemir Could you please re-review this PR? I updated the protocol definition in accordance with our discussion with @sancar and @asimarslan : instead of adding the new field to the existing message, I create the new message for version 2.2.

sancar (approved):

PR #366, opened 2021-01-11
Improve SQL query cancellation on the client

devozerov:
The protocol part of https://github.com/hazelcast/hazelcast/pull/18047

### DIFF
--- a/protocol-definitions/Sql.yaml
+++ b/protocol-definitions/Sql.yaml
@@ -2,9 +2,10 @@ id: 33
 name: Sql
 methods:
   - id: 1
-    name: execute
+    name: execute_reserved
     since: 2.1
     doc: |
+      THIS MESSAGE IS NO LONGER USED BUT KEPT FOR BACKWARD COMPATIBILITY TESTS
       Starts execution of an SQL query.
     request:
       retryable: false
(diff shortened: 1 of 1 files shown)


sancar (approved):

protocol-definitions/Sql.yaml
```diff
@@ -177,14 +177,14 @@ methods:
               ANY(0)
               ROWS(1)
               UPDATE_COUNT(2)
-    response:
-      params:
         - name: queryId
           type: SqlQueryId
           nullable: true
```
sancar, line 182: @devozerov I think this field is mandatory and can not be `nullable`, wdyt ?

sancar (changes requested):

protocol-definitions/Sql.yaml
devozerov: Yes, it was a bug in 4.1, that I copied blindly to the new message. Fixed in https://github.com/hazelcast/hazelcast-client-protocol/pull/366/commits/993d4a89b25971ea9120471674c320d389a4c030

sancar (approved):

devozerov:
The OS part has been merged. @sancar would you mind merging this PR to the protocol repo?

PR #363, opened 2020-12-29
sql return type POC protocol side

sancar:

PR #370, opened 2021-01-24
SQL: New row serialization protocol

devozerov:
Protocol part for https://github.com/hazelcast/hazelcast/pull/18089
Supersedes https://github.com/hazelcast/hazelcast-client-protocol/pull/363


py/__init__.py
```diff
@@ -93,6 +93,10 @@ class PathHolders:
     StringCodec = ImportPathHolder("StringCodec", "protocol.builtin", )
     ListLongCodec = ImportPathHolder("ListLongCodec", "protocol.builtin")
     ListIntegerCodec = ImportPathHolder("ListIntegerCodec", "protocol.builtin")
+    ListCNBooleanCodec = ImportPathHolder("ListCNBooleanCodec", "protocol.builtin")
```
ihsandemir, line 96: does python implement sql yet? what is this change for?

ts/__init__.py
```diff
@@ -156,6 +160,10 @@ class PathHolders:
     'EndpointQualifier': [PathHolders.EndpointQualifier, PathHolders.EndpointQualifierCodec],
     'List_Long': [PathHolders.Long, PathHolders.ListLongCodec],
     'List_Integer': [PathHolders.ListIntegerCodec],
+    'ListCN_Boolean': [PathHolders.ListCNBooleanCodec],
```
ihsandemir, line 163: does nodejs implement sql yet? what is this change for?

py/__init__.py
devozerov: I removed changes from the Python and NodeJS parts. 

However,  perhaps I missed a couple of other things. Should I need the new CN codecs to `util.py`, `protocol-schema.json`, and `custom-codec-schema.json`?

ts/__init__.py
devozerov: Removed.

py/__init__.py
mdumandag: Well, this is not needed until they are being used in the protocol definitions as parameter types but to be future-proof, we might add these new types to `binary/__init__.py`, `binary/__util__.py` and schema files. I believe each language could add it to its `lang/__init__.py` file because each one uses a different style for this file for their own needs. Maybe we can just add them to `java/__init__.py`

Leaving it as it is is also fine for me

ihsandemir (approved):

devozerov:
The IMDG part has been merged to `master`. Could you please merge this PR as well?

py/__init__.py
devozerov: @mdumandag I think, we can expose these classes a little bit later, once I finish with the rest SQL types (double, time, etc.), which would require more custom codecs.

PR #372, opened 2021-01-27
Add perEntryStatsEnabled to MapConfig

ahmetmircik:
depends on https://github.com/hazelcast/hazelcast/pull/18093

### DIFF
--- a/protocol-definitions/DynamicConfig.yaml
+++ b/protocol-definitions/DynamicConfig.yaml
@@ -880,7 +880,7 @@ methods:
         - name: perEntryStatsEnabled
           type: boolean
           nullable: false
-          since: 2.0
+          since: 2.2
           doc: |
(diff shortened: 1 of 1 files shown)


protocol-definitions/DynamicConfig.yaml
```diff
@@ -877,6 +877,13 @@ methods:
           doc: |
             metadata policy configuration for the supported data types. Valid values
             are {@code CREATE_ON_UPDATE} and {@code OFF}
+        - name: perEntryStatsEnabled
+          type: boolean
+          nullable: false
+          since: 2.0
```
sancar, line 883: @ahmetmircik This should be 2.2

sancar (approved):

protocol-definitions/DynamicConfig.yaml
ahmetmircik: fixed

PR #374, opened 2021-02-17
CSharp - Activate AtomicXxxCodec, deal with Migration

zpqrtbnk:
Update CSharp code so that [this PR](https://github.com/hazelcast/hazelcast-csharp-client/pull/423) can implement `IAtomicLong`.


cs/build.sh
```diff
@@ -4,29 +4,36 @@ PROTOCOL_DIR="$( dirname $SCRIPT_DIR )"
 
 pushd $PROTOCOL_DIR
 
+# generate codecs
```
asimarslan, line 7: This build.sh looks like a workaround to `generator.py`. 
Can we move it to `generator.py` so that all other clients may benefit?

zpqrtbnk: looking into it - might kill it instead as it's rather CS specific

asimarslan (approved):

asimarslan (approved):

PR #375, opened 2021-02-24
Refactor Go Generator

yuce:


puzpuzpuz:
This PR seems to be stale while the usual approach is to merge client protocol PR along with the client one (https://github.com/hazelcast/hazelcast-go-client/pull/552 in this case, I guess).

@yuce could you update it?

PR #378, opened 2021-04-15
Adding messages for new serialization format

sancar:


protocol-definitions/Client.yaml
```diff
@@ -886,3 +886,58 @@ methods:
           since: 2.2
           doc: |
             true if the listener existed and was removed, false otherwise.
+  - id: 19
+    name: sendSchema
+    since: 2.3
+    doc: |
+      Sends a schema to cluster
+    request:
+      retryable: false
```
ihsandemir, line 895: Normally all these new messages can be retryable logically, but is there a problem if the retry takes a long time?

ihsandemir (approved):

PR #380, opened 2021-06-16
Add SQL-related codec support for the Python client

mdumandag:
Excluded SQL (apart from the reserved and unused ones) codecs from
the ignorelist and updated the templates with the necessary changes.

Since some classes used in custom codecs are private on the client,
an extra care was needed.

### DIFF
--- a/py/__init__.py
+++ b/py/__init__.py
@@ -162,7 +162,6 @@ import_paths = {
     "List_String": [PathHolders.ListMultiFrameCodec, PathHolders.StringCodec],
     "List_Data": [PathHolders.ListMultiFrameCodec, PathHolders.DataCodec],
     "ListCN_Data": [PathHolders.ListMultiFrameCodec, PathHolders.DataCodec],
-    "List_ListCN_Data": [PathHolders.ListMultiFrameCodec, PathHolders.ListCNDataCodec],
     "List_MemberInfo": [PathHolders.ListMultiFrameCodec, PathHolders.MemberInfoCodec],
     "List_DistributedObjectInfo": [PathHolders.ListMultiFrameCodec, PathHolders.DistributedObjectInfoCodec],
     "List_StackTraceElement": [PathHolders.ListMultiFrameCodec, PathHolders.StackTraceElementCodec],
@@ -237,7 +236,6 @@ _py_types = {
     "List_String",
     "List_StackTraceElement",
     "ListCN_Data",
-    "List_ListCN_Data",
     "List_SqlColumnMetadata",
 
     "EntryList_UUID_Long",
--- a/py/__init__.py
+++ b/py/__init__.py
@@ -4,7 +4,7 @@ import re
 
 def py_types_encode_decode(t):
     if t not in _py_types:
-        raise NotImplementedError("Missing type Mapping")
+        raise NotImplementedError("Missing type Mapping for %s" % t)
 
 
 _pattern1 = re.compile("(.)([A-Z][a-z]+)")
@@ -21,6 +21,15 @@ def py_param_name(type_name):
 
 def py_get_import_path_holders(param_type):
     return import_paths.get(param_type, [])
+
+
+_private_custom_types = ("SqlError", "SqlQueryId")
+
+
+def py_custom_type_name(name):
+    if name in _private_custom_types:
+        return "_" + name
+    return name
 
 
 py_ignore_service_list = {
@@ -54,7 +63,8 @@ py_ignore_service_list = {
     "Queue.drainTo",
     "ReplicatedMap.addNearCacheEntryListener",
     "ScheduledExecutor",
-    "Sql",
+    "Sql.execute_reserved",
+    "Sql.fetch_reserved",
     "Topic.publishAll",
     "TransactionalMap.containsValue",
     "XATransaction",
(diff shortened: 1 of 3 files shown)


py/__init__.py
```diff
@@ -54,7 +63,8 @@ def py_get_import_path_holders(param_type):
     "Queue.drainTo",
     "ReplicatedMap.addNearCacheEntryListener",
     "ScheduledExecutor",
-    "Sql",
+    "Sql.execute_reserved",
+    "Sql.fetch_reserved",
```
srknzl, line 67: why these needed? I deleted them and I think they are still not generated

py/__init__.py
```diff
@@ -144,9 +162,11 @@ class PathHolders:
     "List_String": [PathHolders.ListMultiFrameCodec, PathHolders.StringCodec],
     "List_Data": [PathHolders.ListMultiFrameCodec, PathHolders.DataCodec],
     "ListCN_Data": [PathHolders.ListMultiFrameCodec, PathHolders.DataCodec],
+    "List_ListCN_Data": [PathHolders.ListMultiFrameCodec, PathHolders.ListCNDataCodec],
```
srknzl, line 165: Is this needed somewhere? I don't think this is used in SQL

py/__init__.py
```diff
@@ -122,6 +133,13 @@ class PathHolders:
                                               "protocol.codec.custom.endpoint_qualifier_codec")
     RaftGroupId = ImportPathHolder("RaftGroupId", "protocol")
     RaftGroupIdCodec = ImportPathHolder("RaftGroupIdCodec", "protocol.codec.custom.raft_group_id_codec")
+    SqlQueryId = ImportPathHolder("_SqlQueryId", "sql")
+    SqlQueryIdCodec = ImportPathHolder("SqlQueryIdCodec", "protocol.codec.custom.sql_query_id_codec")
+    SqlColumnMetadata = ImportPathHolder("SqlColumnMetadata", "sql")
+    SqlColumnMetadataCodec = ImportPathHolder("SqlColumnMetadataCodec", "protocol.codec.custom.sql_column_metadata_codec")
+    SqlError = ImportPathHolder("_SqlError", "sql")
+    SqlErrorCodec = ImportPathHolder("SqlErrorCodec", "protocol.codec.custom.sql_error_codec")
+    SqlPageCodec = ImportPathHolder("SqlPageCodec", "protocol.builtin")
```
srknzl, line 142: Why don't we have SqlPage import Path holder

py/__init__.py
```diff
@@ -168,7 +188,11 @@ class PathHolders:
     "PagingPredicateHolder": [PathHolders.PagingPredicateHolder, PathHolders.PagingPredicateHolderCodec],
     "EndpointQualifier": [PathHolders.EndpointQualifier, PathHolders.EndpointQualifierCodec],
     "Map_EndpointQualifier_Address": [PathHolders.MapCodec, PathHolders.EndpointQualifierCodec,
-                                      PathHolders.AddressCodec]
+                                      PathHolders.AddressCodec],
+    "SqlQueryId": [PathHolders.SqlQueryId, PathHolders.SqlQueryIdCodec],
+    "SqlColumnMetadata": [PathHolders.SqlColumnMetadata, PathHolders.SqlColumnMetadataCodec],
+    "SqlError": [PathHolders.SqlError, PathHolders.SqlErrorCodec],
+    "SqlPage": [PathHolders.SqlPageCodec],
```
srknzl, line 195: Don't we need `PathHolders.SqlPage` here after adding a SqlPage PathHolders?

py/__init__.py
```diff
@@ -23,6 +23,15 @@ def py_get_import_path_holders(param_type):
     return import_paths.get(param_type, [])
 
 
+_private_custom_types = ("SqlError", "SqlQueryId")
+
+
+def py_custom_type_name(name):
+    if name in _private_custom_types:
+        return "_" + name
+    return name
```
srknzl, line 32: Did not we need this one before? Is this the first time a private custom type is introduced?

py/__init__.py
mdumandag: For your case, they won't be generated because there are missing type mappings for the TypeScript codecs (`List_ListCN_Data`). If you add it, you will see that they will be generated.

You have to add them here to be sure that they won't be generated

py/__init__.py
mdumandag: Because this is a built-in codec, not a custom one. We don't need `SqlPage` in this case

py/__init__.py
mdumandag: No, as this is a built-in codec

py/__init__.py
mdumandag: Yes, this is the first time

py/__init__.py
mdumandag: Good catch, removed it

py/__init__.py
srknzl: no i tried python codecs and they are not generated

py/__init__.py
srknzl: I wanted to be sure we are on the same thing, I am talking about adding a seperate import path holder for SqlPage, not adding SqlPage inside SqlPageCodec import path holder

mdumandag: It won't be used anywhere, so no point in adding it

srknzl (approved):

PR #385, opened 2021-07-16
Move Jet protocol definitions to this repository

mdumandag:
Apart from moving the definitions, ignored Jet from id validation
checks.

Jet was once living in its repository and given an id of `254`, which
does not fit into valid id range. We could have changed the id now,
but that means we won't be able to read messages from old Jet clusters,
which we probably don't want.


sancar (approved):

frant-hartm (approved):
Generates the same result on my machine as in https://github.com/hazelcast/hazelcast/pull/19118. So that's all good.

There is an empty file `codec_generator.py`, not sure what for.

PR #387, opened 2021-07-30
Include the coordinator UUID in JetSubmitJob

viliam-durina:
Related to hazelcast/hazelcast#19230

### DIFF
--- a/protocol-definitions/Jet.yaml
+++ b/protocol-definitions/Jet.yaml
@@ -24,9 +24,9 @@ methods:
           nullable: true
           since: 2.0
           doc: ''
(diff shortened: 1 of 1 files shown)


sancar:
@viliam-durina we don't normally change a type in the protocol because it is not backward compatible(We are only allowed to adda new field). Can you remind me that what is the situation with jet ? Should 4.0 jet clients work with 5.0 cluster ?

sancar (approved):

viliam-durina:
The related PR hazelcast/hazelcast#19230 was merged

PR #394, opened 2021-09-13
Update python codecs

mdumandag:
- Removed the aggregator and projection related codecs from the ignore
list as we have implemented these features.
- Changed the folder of the EndpointQualifier to its updated place.


srknzl:
Regarding deletions in ignore list: If you implemented them previously and change ignore list now, how do they work? I guess you added the codecs to python client via generating on your local

mdumandag:
Yes, basically, the community user who implemented these features did that. I forget to inform him to create PRs here. So, I am doing it now

srknzl (approved):

PR #400, opened 2021-09-23
MC-635 adding operation resetAgeStatistics

erosb:
Adds a simple operation for resetting the minAge, maxAge and averageAge metrics of a queue. 

Question to the reviewers: what should be the `since` version of new operations?

Platform-side PR: https://github.com/hazelcast/hazelcast/pull/19625

### DIFF
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -846,7 +846,7 @@ methods:
       partitionIdentifier: -1
     response: {}
   - id: 33
-    name: resetAgeStatistics
+    name: resetQueueAgeStatistics
     since: 2.4
     doc: Resets the max-age, min-age and average-age metrics of the queue
     request:
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -845,3 +845,17 @@ methods:
       retryable: false
       partitionIdentifier: -1
     response: {}
+  - id: 33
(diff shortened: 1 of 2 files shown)


mdumandag:
This operation will be used by the management center only, right? I feel like `MC` service would be a better place for it, but not %100 sure. @sancar WDYT?

sancar:
https://github.com/hazelcast/hazelcast-client-protocol/blob/master/protocol-definitions/MC.yaml#L157
It looks like there are similar usages on the MC.yaml 
I would also prefer to keep them MC.yaml.

erosb:
Hello @sancar @mdumandag I moved the new operation to MC.yaml.
One more question: it is tagged with `since: 2.4` . Is it a correct version number?

mdumandag:
@erosb thanks. Yes, that is the correct version number

protocol-definitions/MC.yaml
```diff
@@ -845,3 +845,17 @@ methods:
       retryable: false
       partitionIdentifier: -1
     response: {}
+  - id: 33
+    name: resetAgeStatistics
```
mdumandag, line 849: Maybe rename it to `resetQueueAgeStatistics` as this is under the new namespace that is not related to the queue?

erosb:
run-lab-run

mdumandag (approved):

PR #412, opened 2022-01-27
MC-1268 Adding operations for configuration reload & update

erosb:
Fixes [MC-1268](https://hazelcast.atlassian.net/browse/MC-1268)

Adds `reloadConfig` and `memberConfig` operations. These are remote interfaces for the `ConfigurationService#update()` methods implemented as part of [HZ-760](https://hazelcast.atlassian.net/browse/HZ-760).

OS PR: https://github.com/hazelcast/hazelcast/pull/20529
EE PR: https://github.com/hazelcast/hazelcast-enterprise/pull/4621

### DIFF
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -859,3 +859,37 @@ methods:
          since: 2.4
          doc: Name of the Queue for which the age-related statistics should be reset to 0
     response: {}
+  - id: 34
+    name: reloadConfig
+    since: 2.4
+    doc: Reloads the configuration on the member from its config file, applies (merges) the new config to the current config
+    request:
+      retryable: false
+      partitionIdentifier: -1
+    response:
+      params:
+        - name: configUpdateResult
+          type: ConfigUpdateResult
+          since: 2.4
+          nullable: false
+          doc: the result of the configuration update
+  - id: 35
+    name: updateConfig
+    since: 2.4
+    doc: Reloads the configuration on the member from its config file, applies (merges) the new config to the current config
+    request:
+      retryable: false
+      partitionIdentifier: -1
+      params:
+       - name: configXml
+         type: String
+         nullable: false
+         since: 2.4
+         doc: the new configuration XML to be applied to the current configuration
+    response:
+      params:
+        - name: configUpdateResult
+          type: ConfigUpdateResult
+          since: 2.4
+          nullable: false
+          doc: the result of the configuration update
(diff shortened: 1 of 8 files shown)


protocol-definitions/MC.yaml
```diff
@@ -859,3 +859,37 @@ methods:
...
+      params:
+        - name: configUpdateResult
+          type: ConfigUpdateResult
+          since: 2.4
+          nullable: false
+          doc: the result of the configuration update
+  - id: 35
+    name: updateConfig
+    since: 2.4
+    doc: Reloads the configuration on the member from its config file, applies (merges) the new config to the current config
```
ramizdundar, line 879: minor: Reloads -> Updates

protocol-definitions/custom/Custom.yaml
```diff
@@ -996,3 +996,27 @@ customTypes:
...
+    returnWithFactory: true
+    params:
+      - name: configSectionName
+        type: String
+        since: 2.4
+        nullable: false
+      - name: configName
+        type: String
+        since: 2.4
+        nullable: false
```
ramizdundar, line 1022: Name can be null.

protocol-definitions/MC.yaml
```diff
@@ -859,3 +859,37 @@ methods:
...
          doc: Name of the Queue for which the age-related statistics should be reset to 0
     response: {}
+  - id: 34
+    name: reloadConfig
+    since: 2.4
+    doc: Reloads the configuration on the member from its config file, applies (merges) the new config to the current config
+    request:
+      retryable: false
+      partitionIdentifier: -1
+    response:
```
promanenko, line 869: do we need to return a result of the update here?
We are going to get status events anyway, may be it makes sense to have this process async

protocol-definitions/MC.yaml
```diff
@@ -859,3 +859,37 @@ methods:
...
+      params:
+        - name: configUpdateResult
```
promanenko, line 871: we need to get information about the error in case of configuration update failure

erosb: If we make this async (as per your other comment), then the failure reason will be included in the `ConfigUpdateFailedEvent` we receive on the MC side, won't it?

promanenko: right

erosb: ok thanks I will rework the PRs to make this operation async

mdumandag (approved):
Looks OK from the protocol's point of view.

We can merge this once the Hazelcast PRs are merged

promanenko (approved):

PR #427, opened 2022-07-12
Add SQL query information to Jet admin API

krzysztofslusarski:
Fixes https://github.com/hazelcast/hazelcast/issues/20372


yuce (dismissed):

protocol-definitions/Jet.yaml
```diff
@@ -286,3 +286,18 @@ methods:
...
+    since: 2.5
+    doc: ''
+    request:
+      retryable: true
+      partitionIdentifier: -1
+      params: []
+    response:
+      params:
+        - name: response
+          type: HazelcastJsonValue
```
mdumandag, line 300: As far as I understand, this actually should be `List_JobAndSqlSummary`, where `JobAndSqlSummary` would be a new custom type definition. 

I don't think it is okay to return the response as JSON here. It might be flexible and easy to use, but we need to maintain backward/forward compatibility of the protocol through this repository, and JSON could change anytime without us noticing anything

krzysztofslusarski: Done

protocol-definitions/custom/Custom.yaml
```diff
@@ -1056,3 +1056,63 @@ customTypes:
         type: DiskTierConfig
         nullable: false
         since: 2.5
+  - name: JobStatus
```
mdumandag, line 1059: Is there a reason we have this as a custom type definition? I believe we can make it an `int` field of `JobAndSqlSummary`

binary/util.py
```diff
@@ -300,6 +301,12 @@ def encode_string_frame(client_message, value=None):
     def encode_data_frame(client_message):
         client_message.add_frame(Frame(reference_objects.DATA))
 
+    @staticmethod
+    def encode_json(client_message):
+        client_message.add_frame(BEGIN_FRAME)
+        client_message.add_frame(Frame("{'value': ''}".encode('utf-8')))
+        client_message.add_frame(END_FRAME)
+
```
mdumandag, line 309: I guess we won't need those anymore

binary/util.py
```diff
@@ -236,7 +236,8 @@ def __init__(self, encoder):
             'List_Data': partial(self.encode_multi_frame_list, encoder=self.encode_data_frame),
             'List_ScheduledTaskHandler': partial(self.encode_multi_frame_list, encoder=self.encoder.custom_type_encoder
                                                  .encoder_for('ScheduledTaskHandler')),
-            'SqlPage': partial(self.encode_sqlpage)
+            'SqlPage': partial(self.encode_sqlpage),
+            'HazelcastJsonValue': partial(self.encode_json)
```
mdumandag, line 240: I guess we won't need those anymore

binary/util.py
krzysztofslusarski: I believe we need this for proper usage of ```HazelcastJsonValue```. That type was introduced in ```2.4```, but nobody uses it. When I tried to use it the generated tests failed.

binary/util.py
krzysztofslusarski: I believe we need this for proper usage of ```HazelcastJsonValue```. That type was introduced in ```2.4```, but nobody uses it. When I tried to use it the generated tests failed.

protocol-definitions/custom/Custom.yaml
mdumandag: Yes, it will stay as an enum in the Java code. We will make it an `int` parameter here, and add a method to https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/codec/builtin/FixedSizeTypesCodec.java for it to do the conversion

mdumandag: See this for example https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/codec/builtin/FixedSizeTypesCodec.java#L120

krzysztofslusarski: Done

binary/util.py
```diff
@@ -444,13 +451,16 @@ def encoder_for(self, param_type):
...
     'List_SqlColumnMetadata': 'aListOfSqlColumnMetadata',
+    'List_JobAndSqlSummary': 'aListJobAndSqlSummary',
     'MergePolicyConfig': 'aMergePolicyConfig',
     'CacheConfigHolder': 'aCacheConfigHolder',
     'AnchorDataListHolder': 'anAnchorDataListHolder',
     'PagingPredicateHolder': 'aPagingPredicateHolder',
     'SqlQueryId': 'anSqlQueryId',
     'SqlError': 'anSqlError',
     'SqlColumnMetadata': 'anSqlColumnMetadata',
+    'JobAndSqlSummary': 'anJobAndSqlSummary',
```
mdumandag, line 462:
```suggestion
    'JobAndSqlSummary': 'aJobAndSqlSummary',
```
minor

mdumandag (approved):
approving with a minor comment

mdumandag (approved):

yuce (approved):

PR #428, opened 2022-07-22
[HZ-1272] B-Tree Index Configuration

TomaszGaweda:
This PR adds B-Tree index configuration, added in https://github.com/hazelcast/hazelcast/pull/21820


protocol-definitions/custom/Custom.yaml
```diff
@@ -194,6 +194,10 @@ customTypes:
         type: BitmapIndexOptions
         nullable: true
         since: 2.0
+      - name: btreeIndexConfig
+        type: BTreeIndexConfig
+        nullable: true
+        since: 2.0
```
ufukyilmaz, line 200:
```suggestion
        since: 2.5
```

ufukyilmaz:
Could you add `BTreeIndexConfig` to https://github.com/hazelcast/hazelcast-client-protocol/blob/8db5d9828132ea60d6d8755bf7e67058c9191b91/binary/__init__.py#L65-L91? Then, please regenerate the codecs again

mdumandag (commented):
LGTM overall. Just have one question: Are these new fields really nullable? In the past, we made a mistake by defining some non-nullable fields as nullable and I would like to make sure that we don't make the same mistake here.

TomaszGaweda:
if user do not provide values, the default configuration is taken

mdumandag:
@TomaszGaweda then, can the data carried over with this protocol be ever null? I guess we won't let user set null values for those fields, and if not set, we will use the default configuration, which I assume is also not null. Is this correct?

TomaszGaweda:
Ahh, that way... Didn't think about it. Ok then it should be non-null, I will make a change rn

protocol-definitions/custom/Custom.yaml
```diff
@@ -194,6 +194,10 @@ customTypes:
         type: BitmapIndexOptions
         nullable: true
         since: 2.0
+      - name: bTreeIndexConfig
+        type: BTreeIndexConfig
+        nullable: true
```
mdumandag, line 199: should we make this non-nullable as well?

TomaszGaweda: same as BitmapIndexOptions, user may give null value

mdumandag (approved):

ufukyilmaz (approved):
LGTM. Thanks for your effort.

emreyigit (approved):

TomaszGaweda:
Thanks for the approvals, I've merged the core PR, can you please merge this one as well? :)

PR #436, opened 2022-09-21
MC-1245 Add clusterId to MCGetClusterMetadata endpoint

barisbasaran:
Adding clusterId to MCGetClusterMetadata endpoint. This will be used by Management Center to report the values to Phone Home.
Jira ticket -> [MC-1245 ClusterID in the MC Phonehomes](https://hazelcast.atlassian.net/browse/MC-1245)

Hazelcast PR: https://github.com/hazelcast/hazelcast/pull/22280

### DIFF
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -417,7 +417,7 @@ methods:
           doc: |
             Cluster-wide time in milliseconds.
         - name: clusterId
-          type: String
+          type: UUID
           nullable: false
           since: 2.5
           doc: |
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -416,6 +416,12 @@ methods:
           since: 2.0
           doc: |
             Cluster-wide time in milliseconds.
+        - name: clusterId
+          type: String
+          nullable: false
+          since: 2.5
+          doc: |
+            Cluster ID.
   - id: 15
     name: shutdownCluster
     since: 2.0


mdumandag (approved):

protocol-definitions/MC.yaml
```diff
@@ -416,6 +416,12 @@ methods:
           since: 2.0
           doc: |
             Cluster-wide time in milliseconds.
+        - name: clusterId
```
ihsandemir, line 419: `ClusterId` is already retrieved by the client during authentication response and its type is UUID. See [here](https://github.com/hazelcast/hazelcast-client-protocol/blob/master/protocol-definitions/Client.yaml#L108) Why don't you just use it?

barisbasaran: I guess it's because it's a place where we don't have client authentication information, see https://github.com/hazelcast/management-center/blob/ab17d533efdb60de83ee4b38897d60a381a8285b/src/main/java/com/hazelcast/webmonitor/service/telemetry/ClusterUsageTracker.java#L76

barisbasaran:
hi @zhmaeff @promanenko 
For some reason github does not allow me to add you as reviewers, can you guys also check the PR?

protocol-definitions/MC.yaml
promanenko: @ihsandemir is `clusterId` exposed from the Client? It seems it is not

mdumandag: No, it is only stored in the TcpClientConnectionManager, and we don't expose it. It should be possible to expose it, but I am not sure if it is worth it

barisbasaran: I also checked `ClientAuthenticationCustomCodec#clusterId` in Platform, but the field doesn't seem to be passed anywhere

promanenko: I think it shouldn't be exposed as it has no meaning for customers. Therefore exposure in the private MC operation is a better option

ihsandemir: yes, it is not in the public API. Do you need the `clusterId` of the cluster that the MC client is connected and as seen by MC client? What about the clusterId of the other connected real clients?

barisbasaran: We need it only for MC, we are going to pass it to Phone Home to store in `callHome::mc_info` table.

barisbasaran: FYI I changed the type to UUID if that represents the field better

barisbasaran: to wrap up, shall we approve and move forward with this if we don't have any better option?

mdumandag: I think the current approach is the simplest and the best one. @ihsandemir WDYT?

mdumandag (approved):

protocol-definitions/MC.yaml
ihsandemir: One more remark, if we obtain it from the client itself, we can get this stat even for older version servers. See [this](https://github.com/hazelcast/hazelcast/pull/22280#discussion_r976701223)

barisbasaran: I assume we will miss 5.2 release if we take the way to expose it in client

barisbasaran: Since today we have to wrap up everything for 5.2.0

ihsandemir: If it is urgent for 5.2 release, we can not do the client changes that quick. Let's proceed.

ihsandemir (approved):

PR #437, opened 2022-09-21
[API-1552] Add contains serialized data in requests flag to client messages

mdumandag:
This PR adds the aforementioned flag to the client messages.

### DIFF
--- a/util.py
+++ b/util.py
@@ -170,9 +170,6 @@ def generate_data_containing_requests_lookup_table(services, custom_services):
             if custom_type_contains_serialized_data(type_name):
                 types_containing_serialized_data.add(type_name)
                 return True
-            else:
-                types_not_containing_serialized_data.add(type_name)
-                return False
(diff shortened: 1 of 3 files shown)


util.py
```diff
@@ -123,8 +124,88 @@ def filter_new_params(params, version):
...
+                types_not_containing_serialized_data.add(type_name)
+                return False
+
+        if type_name in custom_types:
+            if custom_type_contains_serialized_data(type_name):
+                types_containing_serialized_data.add(type_name)
+                return True
+            else:
+                types_not_containing_serialized_data.add(type_name)
+                return False
```
srknzl, line 175: Can't you remove these else blocks because at the end of the method we will add return false and call `types_not_containing_serialized_data.add(type_name)` ?

srknzl (approved):

ihsandemir (approved):

PR #441, opened 2022-11-16
Partition-aware client protocol changes [HZ-1657]

ivanthescientist:


protocol-definitions/Sql.yaml
```diff
@@ -222,6 +222,12 @@ methods:
           since: 2.5
           doc: |
             Is the result set unbounded.
+        - name: partitionArgumentIndex
+          type: int
+          nullable: false
+          since: 2.5
```
k-jamroz, line 228: we had a [discussion](https://hazelcast.slack.com/archives/C034YQBTG/p1668784483838309) about using Hazelcast Version for `since` but without final conclusion yet. Anyway, I think it shoud not be 2.5 which is version for 5.2

mdumandag (approved):

k-jamroz (approved):

ihsandemir (approved):

PR #442, opened 2022-11-19
CSharp Updates

zpqrtbnk:
Update to the protocol:
* Add some general functions that we require for C# (filtering of codec properties...)
* Update C# templates, definitions etc to align with the latest `master` protocol


emreyigit (dismissed):

emreyigit (dismissed):

util.py
```diff
@@ -699,10 +723,13 @@ def create_environment(lang, namespace):
...
     env.globals["custom_type_name"] = language_specific_funcs["custom_type_name"][lang]
+    env.globals["custom_codec_param_name"] = language_specific_funcs["custom_codec_param_name"][lang]
     env.globals["get_size"] = get_size
     env.globals["is_trivial"] = is_trivial
-    env.globals["get_import_path_holders"] = language_specific_funcs["get_import_path_holders"][
-        lang
-    ]
+    env.globals["get_import_path_holders"] = language_specific_funcs["get_import_path_holders"][lang]
+    env.globals["copyright_year"] = date.today().year
+    env.globals["protocol_commit"] = os.popen("git rev-parse --short HEAD").readlines()[0].strip()
```
yuce, line 731: I think this wouldn't work if this repo was downloaded (instead of cloned), or `git` is not available in the `PATH`. Should we set`protocol_commit` to some default if an exception is thrown here?

yuce: The correct usage of `os.poopen` would be:
```
with os.popen("git rev-parse --short HEAD") as f:
  env.globals["protocol_commit"] = f.readlines()[0].strip()
```

zpqrtbnk: can do, pushed a fix

emreyigit (approved):

yuce (approved):
Looks good!

emreyigit (approved):

PR #444, opened 2022-12-01
Replace since versions with the Hazelcast versions

mdumandag:
We were using different versioning for the protocol, which was started with 2.0 and bumped for each Hazelcast release.

That was causing some confusion around which protocol version to use, especially when the development for the new Hazelcast versions started.

The initial idea of using a separate protocol version was that we might not need to bump the protocol version if there were no changes in the Hazelcast versions. Over time, we see that we are always bumping the protocol version, because small or big, there are always some changes in the Hazelcast protocol-wise.

We have decided to use Hazelcast versions as the since values from now on. This PR updates the protocol definitions with the new since version using the following script.

```python
import os
import re


mappings = {
    "2.0": "4.0",
    "2.0.1": "4.0.1",
    "2.1": "4.1",
    "2.2": "4.2",
    "2.3": "5.0",
    "2.4": "5.1",
    "2.5": "5.2",
    "2.6": "5.3",
}

pattern = re.compile(r"\s*since: (2\.[0-6](\.1)?)")


def scan_and_rewrite(folder):
    with os.scandir(folder) as scan:
        for entry in scan:
            if not entry.is_file():
                scan_and_rewrite(f"{folder}/{entry.name}")
                continue

            with open(entry, "r") as file:
                new_file_path = f"new/{folder}"
                os.makedirs(new_file_path, exist_ok=True)
                with open(f"{new_file_path}/{entry.name}", "w") as new_file:
                    for line in file.read().splitlines():
                        match = pattern.match(line)
                        if match:
                            old_since = match.groups()[0]
                            new_since = mappings[old_since]
                            line = line.replace(f"since: {old_since}", f"since: {new_since}")

                        new_file.write(line + "\n")


scan_and_rewrite("protocol-definitions")
```

The first commit introduces the changes to the protocol definitions using this script, and the other commit(s) introduces changes to the documentation, README, and code generator scripts.


README.md
```diff
@@ -233,15 +233,17 @@ Client protocol can be expanded by adding new
...
-be equal to the current protocol version. 
+* `since` field of the protocol definitions of the newly added parameters, methods, events and custom types should 
+be equal to the Hazelcast version that the codec is being updated/introduced for. 
 * New services should have the id of the 1 + the highest id of the existing services.
 * New methods should come after the existing methods on the protocol definitions and have the id of the 1 + the id 
 of the method that comes before it.
 * New request, response or event parameters should come after the existing parameters on the protocol definitions 
-and they should be in the increasing order of the protocol versions that is 2.1 parameters should follow 
-2.0.1 parameters which should follow 2.0 parameters.
+and they should be in the increasing order of the Hazelcast versions. That is 4.1 parameters should follow
```
srknzl, line 242:
```suggestion
and they should be in the increasing order of the Hazelcast versions. That is, 4.1 parameters should follow 
```

or 

```suggestion
and they should be in the increasing order of the Hazelcast versions. For example, 4.1 parameters should follow 
```

util.py
```diff
@@ -382,86 +382,49 @@ def validate_services(services, schema_path, no_id_check, protocol_versions):
...
                             event["since"],
                             method_name + "#" + event["name"] + "#event",
                             event_params,
-                            protocol_versions,
                         ):
                             valid = False
     return valid
 
 
-def is_semantically_correct_param(version, protocol_versions):
```
srknzl, line 410: Who will check the semantically correctness now? The reviewer?

srknzl (commented):
Looking good, left a minor suggestion and a question

util.py
mdumandag: No, we don't need this anymore. Previously, we were not OK with skipping a protocol version in definitions (you couldn't use 2.8 when there were no 2.7 defined). But, after sticking to Hazelcast versions, we don't do this anymore. If there will be no changes in the 5.4 release, but some additions to the protocol in 5.5, we will now accept the usage of 5.5 in since versions

README.md
mdumandag: ok, I have changed it

srknzl (dismissed):

zpqrtbnk (approved):
looks good to me

srknzl:
We should also update the change in https://github.com/hazelcast/hazelcast-client-protocol/pull/445/files.

mdumandag:
@srknzl @zpqrtbnk can I get your reviews again? Also, can you guys check the Hazelcast PR as well? I have updated that one too with the new additions, and it was missing an auto-generated test file, I added that one as well

README.md
```diff
@@ -221,6 +221,21 @@ convert it to your enum type and construct the object with it.
...
+|2.0.1|4.0.1|
+|2.1|4.1|
+|2.2|4.2|
+|2.3|5.0|
+|2.4|5.1|
+|2.5|5.2|
+|2.6|5.3|
+|2.7|5.4|
+
+
```
srknzl, line 238: If we are using hazelcast version now, why do we have this table? I think we can't change the older released code, that's why? Similarly, why do we have new/protocol-definitions and protocol-definitions?

ihsandemir (changes requested):
Can you please try generating the classes/docs for all the languages including MD and make sure that they all work without error after this change and you can share sample files for each generated language so that we can observe that they are as expected?

PR #446, opened 2022-12-19
Submit Jet job from non-java client client [HZ-1455]

orcunc:
The client protocol needs to support job uploading. So that non-java clients can also upload jet jobs.
For this purpose two new messages are added to client protocol
1. uploadJobMetaData
2. uploadJobMultipart

uploadJobMetaData Message
The upload process starts with uploadJobMetaData. This message contains 
sessionId, jarSize, snapshotName, jobName, mainClass, jobParameters fields. 
The sessionId is used to distinguish between parallel upload operations 
The jarSize is the size in bytes
The snapshotName, jobName,  mainClass, jobParameters are arguments passed when starting the job

Upon reception of uploadJobMetaData message, the server performs some validation rules. If any of them fail, exception is thrown. If all is well, server stores a new entry in the JobUploadStore class

uploadJobMultipart message
The upload process continues with this message. It contains jar's bytes. This message contains 
sessionId, currentPartNumber, totalPartNumber, partData, partSize fields.

The sessionId is explained in the previous message
The currentPartNumber starts from 1 and shows the sequence number of the part. For example 1 of 5 .
The totalPartNumber is the total number of parts of the sequence
The partData is the byted[] containing jar data
The partSize shows how many bytes of the partData byte[] is valid,

Upon reception of uploadJobMultipart, Various checks are performed on the message and current session.  If any of them fail, exception is thrown. If all is well, message is processed.
If it is the first message a new temporary file is created. Then the partSize bytes of partData byte[] is appended to this file
When all the parts are complete, a new job is started using HazelcastBootstrap.executeJar() call. When the job starts, the jar is loaded into memory and it is not possible to delete it anymore, because it is in use.

HazelcastBootstrap was designed to work only by the hz-client. With this PR it is modified to work on the server side.  However it is still a singleton. 

If any exception is thrown by the server or client fails to finish the upload sequence, a timer in JetServiceBackend cleans the expired JobUploadStore items and deletes the temporary file

Breaking changes (list specific methods/types/messages):
* API
* client protocol format

Implementation PR : https://github.com/hazelcast/hazelcast/pull/23212

Checklist:
- [x] Labels (`Team:`, `Type:`, `Source:`, `Module:`) and Milestone set
- [x] Label `Add to Release Notes` or `Not Release Notes content` set
- [x] Request reviewers if possible
- [x] New public APIs have `@Nonnull/@Nullable` annotations


protocol-definitions/Jet.yaml
```diff
@@ -301,3 +301,93 @@ methods:
...
+          nullable: false
+          since: 2.6
+          doc: 'The size of binary data'
+    response:
+      params:
+        - name: response
+          type: boolean
+          nullable: false
+          since: 2.6
+          doc: 'True if processed successfully, false otherwise'
```
frant-hartm, line 393: Missing empty line at the end :-)

frant-hartm (approved):

frant-hartm:
Could you please update the title to follow our conventions?

frant-hartm:
Also you should cross-link the PRs in the description

https://github.com/hazelcast/hazelcast/pull/23212

protocol-definitions/Jet.yaml
orcunc: Added new empty line

orcunc:
> Could you please update the title to follow our conventions?

Change the title to "Submit Jet job from non-java client [HZ-1455]"

orcunc:
> Also you should cross-link the PRs in the description
> 
> [hazelcast/hazelcast#23212](https://github.com/hazelcast/hazelcast/pull/23212)

I have added the link of the java implementation PR to description

mdumandag:
Hi, just a heads up: We have updated the code generator to use 2023 as the copyright year. So, don't forget to rebase your branch with master and regenerate codecs and test files so that they have the correct copyright year in the Hazelcast PR.

frant-hartm (approved):

protocol-definitions/Jet.yaml
```diff
@@ -321,3 +321,98 @@ methods:
...
+          nullable: false
+          since: 2.6
+          doc: 'Arguments to pass to the supplied jar file'
+    response:
+      params:
+        - name: response
+          type: boolean
+          nullable: false
+          since: 2.6
+          doc: 'True if processed successfully, false otherwise'
```
yuce, line 373: Maybe it would be better to have this in the TDD, but how can metadata upload fail and this call returns `false`?

protocol-definitions/Jet.yaml
```diff
@@ -321,3 +321,98 @@ methods:
...
+          nullable: false
+          since: 2.6
+          doc: 'Hexadecimal SHA256 of the message part'
+    response:
+      params:
+        - name: response
+          type: boolean
+          nullable: false
+          since: 2.6
+          doc: 'True if processed successfully, false otherwise'
```
yuce, line 418: I guess an error about why the upload failed (e.g., hash doesn't match, no space left on storage) will still be sent, won't it?

protocol-definitions/Jet.yaml
OrcunColak: This call returns true it is succeeds, all other errors are propagated with exception.

protocol-definitions/Jet.yaml
OrcunColak: This call returns true it is succeeds, all other errors are propagated with exception.

yuce (approved):
Thanks, looks good except the message responses. I believe that will be addressed in another PR.

mdumandag (approved):

PR #447, opened 2022-12-23
Add job status listener [HZ-1646]

burakgok:
For hazelcast/hazelcast#23193 and hazelcast/hazelcast-enterprise#5707

PR #450, opened 2023-01-30
Add support for changing JobConfig of a suspended job [HZ-1650]

burakgok:
For hazelcast/hazelcast#23523

### DIFF
--- a/protocol-definitions/Jet.yaml
+++ b/protocol-definitions/Jet.yaml
@@ -416,3 +416,28 @@ methods:
           nullable: false
           since: 2.6
           doc: 'True if processed successfully, false otherwise'
+  - id: 21
+    name: updateJobConfig
+    since: 2.6
+    doc: ''
+    request:
+      retryable: true
+      partitionIdentifier: -1
+      params:
+        - name: jobId
+          type: long
+          nullable: false
+          since: 2.6
+          doc: ''
+        - name: deltaConfig
+          type: Data
+          nullable: false
+          since: 2.6
+          doc: ''
+    response:
+      params:
+        - name: updatedConfig
+          type: Data
+          nullable: false
+          since: 2.6
+          doc: ''


protocol-definitions/Jet.yaml
```diff
@@ -321,3 +321,22 @@ methods:
...
+        - name: jobId
+          type: long
+          nullable: false
+          since: 2.6
+          doc: ''
+        - name: config
+          type: Data
+          nullable: false
+          since: 2.6
+          doc: ''
```
frant-hartm, line 341: Shouldn't we rather introduce a custom type which can be evolved? Otherwise we end up with a binary blob that can't be changed.

frant-hartm (changes requested):

protocol-definitions/Jet.yaml
burakgok: @hazelcast/apisjava This is in line with `submitJob`, which specifies `jobConfig` as `Data`. How should we proceed? What is the benefit of introducing custom types?

mdumandag: The rule of thumb is that, if the parameter is of a known type we should use a custom type to represent it. The reasoning is what Frantisek wrote. If we make it data, it means it can break backward compatibility when the identified data representation of the parameter changes. 

However, I believe those other data parameters were used because Jet doesn't have the concept of backward compatibility. Do we allow old/new clients to submit jobs to new/old members?

burakgok: No, we don't allow that. A client must have the same minor version as the member to which it is connected, so members will always be able to deserialize `JobConfig`s sent by clients.

viliam-durina:
Closing as the author is off and doesn't allow commits to his branch to finish the work.
Replaced by #455

protocol-definitions/Jet.yaml
```diff
@@ -416,3 +416,28 @@ methods:
           nullable: false
           since: 2.6
           doc: 'True if processed successfully, false otherwise'
+  - id: 21
```
mdumandag, line 419: is there a reason we skip id 19 and 20 and jump from 18 to 21 here?

protocol-definitions/Jet.yaml
```diff
@@ -416,3 +416,28 @@ methods:
...
+    name: updateJobConfig
+    since: 2.6
+    doc: ''
+    request:
+      retryable: true
```
mdumandag, line 424: is this really retryable? is it an idempotent operation?

protocol-definitions/Jet.yaml
```diff
@@ -416,3 +416,28 @@ methods:
...
+      partitionIdentifier: -1
+      params:
+        - name: jobId
+          type: long
+          nullable: false
+          since: 2.6
+          doc: ''
+        - name: deltaConfig
+          type: Data
```
mdumandag, line 433: just to double check here about using data here: we don't care about the backward compatibility for jet, right @viliam-durina?

protocol-definitions/Jet.yaml
mdumandag: I assume this is because we use those ids in https://github.com/hazelcast/hazelcast-client-protocol/pull/447. So, I am resolving this, assuming we will merge that PR before 5.3 as well

protocol-definitions/Jet.yaml
viliam-durina: Yes, to use Jet one has to use the exact same minor version. Otherwise many other things will break.
SQL avoids this problem, you can change the job config with a SQL command.

protocol-definitions/Jet.yaml
viliam-durina: Yes, it is idempotent.

protocol-definitions/Jet.yaml
viliam-durina: Right

mdumandag (approved):

viliam-durina:
Replaced by #457, please close this PR.

PR #453, opened 2023-03-06
Submit job from jar already present on member HZ [1974]

orcunc:
This PR allows the client to execute a jar that is already present on the member.  
A new flag has been added  to job upload metadata message
The method calls now return void instead of boolean since the call was never returning false.


mdumandag (dismissed):

yuce (approved):
Looks good!

protocol-definitions/Jet.yaml
```diff
@@ -334,6 +334,11 @@ methods:
           nullable: false
           since: 2.6
           doc: 'Unique session ID of the job upload request'
+        - name: jarOnMember
+          type: boolean
+          nullable: false
+          since: 2.6
+          doc: 'Flag that indicates that jar should be executed, not uploaded'
```
frant-hartm, line 341:
```suggestion
          doc: 'Flag that indicates that the jar to be executed is already present on the member, and no jar will be uploaded from the client'
```

yuce (approved):

protocol-definitions/Jet.yaml
orcunc: Ok

mdumandag (approved):

PR #456, opened 2023-03-08
AttributePartitioningStrategy config protocol changes

ivanthescientist:
Adds custom type for `PartitioningStrategyConfig` entities as well as field for transferring list of partitioning attribute configs.


viliam-durina (dismissed):

protocol-definitions/DynamicConfig.yaml
```diff
@@ -903,6 +903,12 @@ methods:
           since: 2.5
           doc: |
             Tiered-Store configuration
+        - name: partitioningAttributeConfigs
+          type: List_PartitioningAttributeConfig
+          nullable: true
```
mdumandag, line 908: just to double check, as most of the other configs here are not nullable: should we make this nullable?

protocol-definitions/custom/Custom.yaml
```diff
@@ -1144,3 +1144,10 @@ customTypes:
         type: boolean
         nullable: false
         since: 2.6
+  - name: PartitioningAttributeConfig
+    since: 2.6
+    params:
+      - name: attributeName
+        type: String
+        nullable: false
+        since: 2.6
```
mdumandag, line 1153: minor: newline at the end of the file

binary/__init__.py
```diff
@@ -95,6 +95,7 @@
     'MemoryTierConfig',
     'DiskTierConfig',
     'TieredStoreConfig',
+    "PartitioningAttributeConfig",
```
mdumandag, line 98: Minor: I believe this is not needed for now. Maybe we can remove references to non-list field from this PR

protocol-definitions/DynamicConfig.yaml
```diff
@@ -903,6 +903,12 @@ methods:
...
+          since: 2.6
+          doc: |
+            List of Attributes used for creating AttributePartitioningStrategy.
```
mdumandag, line 911:
```suggestion
            List of attributes used for creating AttributePartitioningStrategy.
```

protocol-definitions/DynamicConfig.yaml
ivanthescientist: I'm not sure to be honest, IndexConfigs are optional and its a list too.

srknzl (approved):

PR #458, opened 2023-03-13
Introduce protocol changes for Python compact serialization [API-1564]

mdumandag:


yuce (approved):

py/__init__.py
```diff
@@ -11,10 +11,10 @@ def py_types_encode_decode(t):
 _pattern2 = re.compile("([a-z0-9])([A-Z])")
 
 
-def py_param_name(type_name):
+def py_param_name(type_name, check_for_keywords=True):
```
srknzl, line 14: why do we need this boolean?

mdumandag: await is a keyword in python, and without it, we were generating a codec file named `count_down_latch__await_codec.py` instead of `count_down_latch_await_codec.py` (double vs single underscore before await)

srknzl (approved):

PR #461, opened 2023-03-20
Add to_object function to Python codecs when necessary [API-1964]

mdumandag:
This change was needed to support efficient eager deserialization of the SQL columns of type OBJECT.

I have also refactored the `language_specific_funcs` so that we don't have to put dummy values even for functions that are needed for a single language.

### DIFF
--- a/util.py
+++ b/util.py
@@ -24,12 +24,12 @@ from cpp import (
     cpp_param_name
 )
 from cs import (
-    cs_escape_keyword, 
-    cs_ignore_service_list, 
-    cs_types_decode, 
-    cs_types_encode, 
+    cs_escape_keyword,
+    cs_ignore_service_list,
+    cs_types_decode,
+    cs_types_encode,
     cs_custom_codec_param_name,
-    cs_init_env
+    cs_sizeof
 )
 from java import java_types_decode, java_types_encode
 from md import internal_services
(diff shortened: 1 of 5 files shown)


util.py
```diff
@@ -24,12 +24,12 @@
...
-    cs_ignore_service_list, 
-    cs_types_decode, 
-    cs_types_encode, 
+    cs_escape_keyword,
+    cs_ignore_service_list,
+    cs_types_decode,
+    cs_types_encode,
     cs_custom_codec_param_name,
-    cs_init_env
+    cs_sizeof
```
yuce, line 32: Should the .Net guys review this change?

mdumandag: There is no behaviour change, in the init method they were just adding this function to the env, which we do as well

yuce (approved):

srknzl (commented):
We are passing to_object fn to codecs if one of its response parameters is SqlPage. I think this can be execute and fetch. 

I did not get why we pass to_object fn to SqlPage. Can't we keep a to_object_fn in SqlResult and call it in _IteratorBase._get_current_row after SqlPage.get_value returns a Data.

mdumandag:
@srknzl

1. Yes, if you take a look at the Python PR, these two are the changed codecs.
2. No, that doesn't work, we need to receive "SchemaNotFoundError" while decoding the client message. Doing it on the iterator has the same problems as our old lazily-deserialized list. Plus, we would have to deal with the positions etc. in that case, which would result in a fairly complex solution. This is much simpler, and actually a bit faster, because we now eliminated unnecessary to_object calls for column types that we don't need to deserialize (i.e. everything other than OBJECT)

srknzl (approved):

PR #468, opened 2023-07-27
Add putAllWithMetadata into client protocol [HZ-2583]

frant-hartm:


protocol-definitions/Map.yaml
```diff
@@ -2965,4 +2965,26 @@ methods:
...
+    doc: |
+      Put a batch of entries with metadata into an IMap. This is a client side version of WAN entry replication.
+    request:
+      retryable: false
+      partitionIdentifier: any key belongs to target partition
+      params:
+        - name: name
+          type: String
+          nullable: false
+          since: 2.7
```
frant-hartm, line 2981: Is this version correct?

srknzl: I checked it. 2.6 is for 5.3. Therefore, 2.7 should be fore 5.4 and it is correct.

srknzl (dismissed):

protocol-definitions/Map.yaml
```diff
@@ -2965,4 +2965,26 @@ methods:
...
           doc: |
             the function to apply to each entry.
-    response: {}
\ No newline at end of file
+    response: {}
+  - id: 73
+    name: putAllWithMetadata
+    since: 2.7
+    doc: |
+      Put a batch of entries with metadata into an IMap. This is a client side version of WAN entry replication.
```
ihsandemir, line 2973: may be you can add some more details as in the doc of putAll message (includes some details about partitions etc.).

srknzl (dismissed):

srknzl (approved):

ihsandemir (approved):

PR #470, opened 2023-07-31
Add DynamicConfig.addWanReplicationConfig [HZ-2717]

gbarnett-hz:
These are the associated changes required to support the addition of WAN replication configurations programmatically via the Hazelcast client.

Fixes (API definition): https://github.com/hazelcast/hazelcast/issues/19979 

OSS PR: https://github.com/hazelcast/hazelcast/pull/25118


protocol-definitions/DynamicConfig.yaml
```diff
@@ -1292,3 +1292,37 @@ methods:
           doc: |
             Properties of the data connection configuration.
     response: {}
+  - id: 18
+    name: addWanReplicationConfig
+    since: 2.7
```
gbarnett-hz, line 1297: What version to use?

protocol-definitions/DynamicConfig.yaml
```diff
@@ -1292,3 +1292,37 @@ methods:
...
+    doc: |
+      Adds a WAN replication configuration.
+    request:
+      retryable: false
```
ihsandemir, line 1301: is this operation idempotent? if I call 2 times `ClientDynamicClusterConfig#addWanReplicationConfig` with the same config and name will it change the resulting config?

protocol-definitions/DynamicConfig.yaml
```diff
@@ -1292,3 +1292,37 @@ methods:
...
+      partitionIdentifier: -1
+      params:
+        - name: name
+          type: String
+          nullable: false
+          since: 2.7
+          doc: |
+            WAN replication configuration name.
+        - name: consumerConfig
+          type: Data
```
ihsandemir, line 1311: If you want backward compatibility guarantees, then try to avoid using Data as the type except the user provided fields which we have no control on, e.g. the user key, value for the data. Hence, I would suggest that you encode every field as you can into the message as a field. This will be more work to begin but it pays in the long term. Especially, also consider that this codec may be used by other non-Java clients as well. We can talk about this.

protocol-definitions/DynamicConfig.yaml
gbarnett-hz: Haven't addressed this in latest commit. Will do later.

gbarnett-hz:
@ihsandemir I've converted pretty much everything from `Data` to a custom codec. Could you please check? I had some issue with generating some binary compatibility tests as well -- any tips? If you run `generate.py` you will see the error.

protocol-definitions/DynamicConfig.yaml
gbarnett-hz: @ihsandemir this looks like it could be idempotent as it's predicated on the uniqueness of the configuration name, however it seems like others that follow a similar strategy have been marked as non-retriable. So, I think it's safer  just to follow the same pattern to be on safe side.

binary/__init__.py
```diff
@@ -96,6 +96,17 @@
     'DiskTierConfig',
     'TieredStoreConfig',
     "PartitioningAttributeConfig",
+    "WanConsumerConfigHolder",
+    "WanCustomPublisherConfigHolder",
+    "WanBatchPublisherConfigHolder",
```
srknzl, line 101: DiscoveryConfigHolder is missing. I also realized we didn't regenerate the binary compatibility tests in the main PR.

protocol-definitions/custom/Custom.yaml
```diff
@@ -1151,3 +1151,288 @@ customTypes:
...
+    params:
+      - name: className
+        type: String
+        nullable: false
+        since: 2.7
+      - name: properties
+        type: Map_String_Data
+        nullable: false
+        since: 2.7
+  - name: DiscoveryConfig
```
srknzl, line 1413: I think we should name the following like XXConfigHolder instead of XXXConfig: 

"DiscoveryConfig",
    "DiscoveryStrategyConfig",
    "WanSyncConfig",

binary/util.py
```diff
@@ -487,6 +487,9 @@ def encoder_for(self, param_type):
     'DiskTierConfig': 'aDiskTierConfig',
     'TieredStoreConfig': 'aTieredStoreConfig',
     'List_PartitioningAttributeConfig': 'aListOfPartitioningAttributeConfigs',
+    'WanConsumerConfigHolder': 'aWanConsumerConfigHolder',
+    'List_WanCustomPublisherConfigHolder': 'aListOfWanCustomPublisherConfigsHolders',
+    'List_WanBatchPublisherConfigHolder': 'aListOfWanBatchPublisherConfigHolders',
```
srknzl, line 492: Do you know why we don't need to add an instance here for WanSyncConfig, DiscoveryConfig etc. 🤔 ?

gbarnett-hz: They aren't top-level reference in the `addWanReplicationConfig` API. The `ReferenceObjects` in OS project defines the nested instances.

binary/__init__.py
gbarnett-hz: Everything is there that needs to be, as discussed.

srknzl (approved):
since renames will require OS changes and the PR is already merged, it's ok this way as well (just naming inconsistency)

protocol-definitions/custom/Custom.yaml
gbarnett-hz: Comment captured in later comment.

ihsandemir (approved):

Issue #473, opened 2023-08-15
Document custom types, holder pattern and binary compatibility test process


gbarnett-hz:
For https://github.com/hazelcast/hazelcast/pull/25223 there were a number of scenarios encountered that had no documentation:

- When using types specific to a target language you need to create holder types as an intermediate representation from the protocol types
- Placement of custom types and their mappings in various places
- Binary compatibility tests and how they relate to other projects, e.g. `ReferenceObjects.java` in Hazelcast OS project
- Implied assumptions of generated tests on the domain types

PR #474, opened 2023-08-16
Quick guide for custom types, holder and binary compat tests [HZ-2919]

gbarnett-hz:
Quick example guide for describing the steps required when you start to introduce custom types, holders and the steps necessary for binary compat tests.

Fixes: #473

### DIFF
--- a/custom-types-quick-guide.md
+++ b/custom-types-quick-guide.md
@@ -13,8 +13,6 @@ fictitious method: `addUsefulConfig`.
 _Note_. At the time of writing the highest `id` in `DynamicConfig.yaml` was `18` -- the `id` of the
 last entry, hence the use of `19` for our example.
 
-_Note._
-
 ```yaml
 - id: 19
   name: addUsefulConfig
@@ -193,8 +191,9 @@ instantiate your `ServiceBConfig` type.
 For many examples of how this is done I recommend looking at
 `ClientDynamicClusterConfig#addWanReplicationConfig` which shows mapping of the domain types to
 their respective holder types on the way out to the server; and `AddWanReplicationConfigTask` which
-shows the mapping of holder types to their respective configuration types. This example covers most
-scenarios you would encounter.
+shows the mapping of holder types to their respective configuration types. These can be found in the
+Hazelcast project [here](https://github.com/hazelcast/hazelcast). This example covers most scenarios
+you would encounter.
 
 ## Binary Compatibility Tests
 
--- a/custom-types-quick-guide.md
+++ b/custom-types-quick-guide.md
@@ -1,0 +1,246 @@
+# Custom Types Quick Guide
+
+_Abstract._ We will go over the steps required to extend an existing service with a new API. The
+focus is on covering the details with respect to the client protocol and its related components. We
+will not cover the underlying logic of the API itself. We use Java as the subject language for code
+generation. The [README](README.md) covers the general framework, so please read that first.
+
+## Adding a Service API
+
+Let us take the `DynamicConfig.yaml` service as the baseline for an example. We will add a
+fictitious method: `addUsefulConfig`.
+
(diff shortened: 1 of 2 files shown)


yuce (approved):
Looks good!

custom-types-quick-guide.md
```diff
@@ -0,0 +1,245 @@
...
+      nullable: false
+      since: 2.7
+    - name: addressParser
+      type: Data
+      nullable: false
+      since: 2.7
+```
+
+Open `schema/protocol-schema.json` and append to the array `definitions.param.properties.type.enum`
+the string literals: `ServiceAConfig` and `ServiceBConfig`.
```
srknzl, line 86: I think if we have used a new custom type in a custom type, we would also need to modify custom-codec-schema.json. 

Should we mention it here as well?

custom-types-quick-guide.md
```diff
@@ -0,0 +1,245 @@
...
+- `output/java/hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/ClientCompatibilityNullTest_2_7.java`
+- `output/java/hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/ClientCompatibilityTest_2_7.java`
+- `output/java/hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/MemberCompatibilityNullTest_2_7.java`
+- `output/java/hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/MemberCompatibilityTest_2_7.java`
+
+If you open the test files up in your IDE you will see some reference errors for `aServiceAConfig`
+and `aServiceBConfig`. For example, if you were to open `ClientCompatibilityTest_2_7.java`. You must
+add these to `com.hazelcast.client.protocol.compatibility.ReferenceObjects` in the Hazelcast
+project. If you need a `String` when creating instances of `ServiceAConfig` or `ServiceBConfig` you
+should use `aString` and likewise for other basic types. You should follow this naming strategy when
```
srknzl, line 232: Also as I understand, for any array type it should be a singletonList of that type. Worth adding?

custom-types-quick-guide.md
```diff
@@ -0,0 +1,245 @@
...
+
+The `ServiceBConfigHolder` is the same as `ServiceBConfig` Java class with the exception that we
+have made `addressParser` now the opaque `Data` type.
+
+With the boilerplate in-place we can now add the following to `java/__init__.py`'s
+`_java_types_common` dictionary:
+
+```python
+    "ServiceAConfig": "com.hazelcast.config.ServiceAConfig",
+    "ServiceBConfig": "com.hazelcast.config.ServiceBConfigHolder"
```
srknzl, line 165: this is something we already discussed in your PR, we skipped there not the create friction but for the readme I am note sure. 

so what about naming this ServiceBConfigHolder instead because this is what we did for others in the past.? 

It may make the README a bit confusing but I think we can fix that by describing why we need a holder before using a holder, basically one of the first things to describe.

srknzl (commented):
I am happy with this addition, just a few suggestions I have

custom-types-quick-guide.md
gbarnett-hz: I pointed them to another example for more detail so if they need a load more examples of how to create bits they can look there. The main point of this quick doc is to describe the undocumented pieces that are required in a relatively succinct way.

custom-types-quick-guide.md
srknzl: ok, suggested just to make it comprehensive but not a must

custom-types-quick-guide.md
srknzl: yeah I was talking about describing it earlier if we use holder name because protocol yaml will have it. 

Thing is, the holder is not something user sees anyway, my rename suggestion is just to follow our convention so far.

srknzl (approved):

PR #487, opened 2023-11-08
Add declarative Jet Pipeline definition message [API-2167]

ihsandemir:
Introduced the client message to submit a job using a declarative Jet Pipeline definition together with resources.

### DIFF
--- a/protocol-definitions/Experimental.yaml
+++ b/protocol-definitions/Experimental.yaml
@@ -1,3 +1,31 @@
 id: 253
 name: Experimental
-methods: []
+methods:
+  - id: 1
+    name: pipelineSubmit
+    since: 2.7
+    doc: 'The message is used to transfer the declarative pipeline definition and the related resource files 
+    from client to the server.'
+    request:
+      retryable: true
+      partitionIdentifier: -1
+      params:
+        - name: pipelineDefinition
+          type: String
+          nullable: false
+          since: 2.7
+          doc: 'The definition of the pipeline steps. It currently uses the YAML format.'
+        - name: resourceBundle
+          type: byteArray
+          nullable: true
+          since: 2.7
+          doc: 'This is the zipped file which contains the user project folders and files. For Python project, 
+          it is the Python project files. It is optional in the sense that if the user likes to use a user docker 
+          image with all the resources and project files included, this parameter can be null.'
+    response:
+      params:
+        - name: jobId
+          type: long
+          nullable: false
+          since: 2.7
+          doc: 'This is the unique identifier for the job which is created for this pipeline'
--- a/protocol-definitions/Experimental.yaml
+++ b/protocol-definitions/Experimental.yaml
@@ -10,6 +10,11 @@ methods:
       retryable: true
       partitionIdentifier: -1
       params:
+        - name: jobName
+          type: String
+          nullable: true
+          since: 2.7
+          doc: 'The name of the submitted Job using this pipeline.'
         - name: pipelineDefinition
           type: String
           nullable: false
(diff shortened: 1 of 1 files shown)


protocol-definitions/Experimental.yaml
```diff
@@ -1,3 +1,31 @@
 id: 253
 name: Experimental
-methods: []
\ No newline at end of file
+methods:
+  - id: 1
+    name: pipelineSubmit
```
zpqrtbnk, line 5: The existing JET message is `submitJob` and maybe we want that one to follow a similar name such as `submitYamlJob`? But then.... that existing message is different since it accepts the `jobId` as a parameter. And no binary data.

Also it accepts the `lightJobCoordinator` which we may want to enable here too?

protocol-definitions/Experimental.yaml
```diff
@@ -1,3 +1,31 @@
...
+    request:
+      retryable: true
+      partitionIdentifier: -1
+      params:
+        - name: pipelineDefinition
+          type: String
+          nullable: false
+          since: 2.7
+          doc: 'The definition of the pipeline steps. It currently uses the YAML format.'
+        - name: resourceBundle
```
zpqrtbnk, line 18: Is this for one resource, or several resources? In some situations we may want to provide the job with *two* resources: the "user code" (ie executables) and the "secrets" (needed to connect back to the cluster) which may come from different locations when submitting. Or maybe CLC would aggregate them in one single payload?

What is the max. supported size of the resourceBundle?

zpqrtbnk (commented):
commented

protocol-definitions/Experimental.yaml
ihsandemir: I am OK with naming if nobody rejects it, can you comment in the TDD about this naming suggestion?

ihsandemir: Existing message is different since it is a different message. We return the job Id in the response once the job is started.

Why do you think the `lightJobCoordinator` is needed for our case?

protocol-definitions/Experimental.yaml
ihsandemir: Max. supported via codec will be configurable system wide with a default of 10MB (see the TDD). User can have larger project but then they need to just use their own custom image API but no project code upload.

We discussed multiple resources case, and it was said that we if for example you have multiple Python steps, each may have different module name and they can all be represented inside the project folder, may be a different python file or may be a different subfolder. Hence, the idea is have every code you need inside this zip and let the resources differentiate via the path/file/module names.

Let me know if you see any problem with this approach and we can discuss.

protocol-definitions/Experimental.yaml
zpqrtbnk: Because you may want to submit a job using YAML and set a `lightJobCoordinator` just the way you would do with Java?

protocol-definitions/Experimental.yaml
zpqrtbnk: Problem is I don't understand why we inflict this onto ourselves when the normal Jet submit has already figured it out.

srknzl (commented):
since this is just the yaml content change that's related to the UCD project rather than to the protocol repo, I think @emreyigit should review as the second reviewer.

protocol-definitions/Experimental.yaml
ihsandemir: Please see limitations of a light job at [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/jet/JetService.java#L166). Do we really want to support it at phase 1 from today? If a request comes later, surely we can add. The `lightJobCoordinator` UUID requires that the user uses a specific UUID for a specific member but at Viridian, we manage the cluster and the members, the user normally should not deal with individual members, right? 

do you still think we need this parameter?

protocol-definitions/Experimental.yaml
ihsandemir: We utilize JobConfig.attachFile API. Currently, for Python solution all we need is to attach the user's Python project files, hence, we did not see the need for multiple files. But if there is a need surely, we can consider that. Please provide a valid use case which can not be satisfied with the current approach?

emreyigit (dismissed):

emreyigit (approved):

PR #489, opened 2023-11-30
Add ReplicatedMapEntryView custom type & ReplicatedMap.putAllWithMetadata [HZ-3934]

srknzl:
Mono pr https://github.com/hazelcast/hazelcast-mono/pull/169 

and https://github.com/hazelcast/hazelcast-mono/pull/462/

PR #492, opened 2023-12-18
ReplicatedMap fetchEntryViews API [HZ-3932]

srknzl:
Mono PR: https://github.com/hazelcast/hazelcast-mono/pull/188

This PR extends #489


copy_java_files.sh
```diff
@@ -0,0 +1,11 @@
+# Use this to copy generated java files to hazelcast repo all at once.
```
ihsandemir, line 1: generator.py already has an option where you can specify destination, why do you need this?

protocol-definitions/ReplicatedMap.yaml
```diff
@@ -820,3 +820,113 @@ methods:
             since: 2.0
             doc: |
               Number of entries affected by this event.
+  - id: 19
+    name: putAllWithMetadata
+    since: 2.7
+    doc: |
+      Copies all of the mappings from the specified entry list to this map, including metadata.
+      This method uses ReplicatedRecordStore.putRecords method with version=current version + 1.
```
ihsandemir, line 828: What do you mean by `current version`? How does it work when it is concurrent invocations?

protocol-definitions/ReplicatedMap.yaml
```diff
@@ -820,3 +820,113 @@ methods:
...
+          since: 2.7
+          doc: |
+            version of the replicated map store that stores the data
+    response: {}
+  - id: 20
+    name: fetchEntryViews
+    since: 2.7
+    doc: |
+      Allows iteration over ReplicatedMapEntryViewHolder objects. The changes happened during the iteration may not be included in the iterated EntryViews. This method
+      will throw an exception if there is no replicated record store with the given replicated map name and partition id. This method will consume some memory in the
```
ihsandemir, line 860: how much is this `some memory`? If I have many concurrent requests how is memory usage growing?

copy_java_files.sh
srknzl: yeah apparently it has, I didn't notice. I'll delete the script

protocol-definitions/ReplicatedMap.yaml
srknzl: putRecords is used to migrate all records from one member to another member, every store has some current version that is increased for every modification (I believe). That's why we need the version information here

srknzl: Btw the description is wrong here, we should be just putting records with the provided version. Let me fix here

srknzl: Probably left stale because I based on another protocol PR in this PR

protocol-definitions/ReplicatedMap.yaml
srknzl: Depends on the entry sizes and fetchSize parameter.

frant-hartm (dismissed):

protocol-definitions/ReplicatedMap.yaml
```diff
@@ -852,3 +852,75 @@ methods:
...
     response: {}
+  - id: 20
+    name: fetchEntryViews
+    since: 2.7
+    doc: |
+      Allows iteration over ReplicatedMapEntryViewHolder objects. The changes happened during the iteration may not be included in the iterated EntryViews. This method
+      will throw an exception if there is no replicated record store with the given replicated map name and partition id. This method will consume some memory in the 
+      member with the default timeout of 300 seconds that is reset after each fetch. Sending endEntryViewIteration or timing out will release the resource.
+    request:
+      retryable: true
```
ihsandemir, line 863: You say that this API causes some memory allocation and keeps it at server memory. What happens to that memory on retry since you made it retryable?

protocol-definitions/ReplicatedMap.yaml
```diff
@@ -852,3 +852,75 @@ methods:
...
+          nullable: false
+          since: 2.7
+          doc: |
+            Name of the ReplicatedMap
+        - name: cursorId
+          type: UUID
+          nullable: true
+          since: 2.7
+          doc: |
+            The identifier of the last fetched page. Initially null and then returned by this API.
```
ihsandemir, line 877: How do you guarantee the uniqueness of the UUID? Let's assume I am doing many `fetchEntryViews ` calls(may be million).

protocol-definitions/ReplicatedMap.yaml
```diff
@@ -852,3 +852,75 @@ methods:
...
+        - name: batchSize
+          type: int
+          nullable: false
+          since: 2.7
+          doc: |
+            The maximum number of EntryViews to be returned.
+    response:
+      params:
+        - name: cursorId
+          type: UUID
```
ihsandemir, line 893: is this different from request parameter `cursorId`? Why?

protocol-definitions/ReplicatedMap.yaml
```diff
@@ -852,3 +852,75 @@ methods:
...
+          nullable: false
+          since: 2.7
+          doc: |
+            A unique string which is used to fetch new pages.
```
ihsandemir, line 897: it is not `String`!!! UUID

protocol-definitions/ReplicatedMap.yaml
```diff
@@ -852,3 +852,75 @@ methods:
...
+        - name: entryViews
+          type: List_ReplicatedMapEntryViewHolder
+          nullable: true
+          since: 2.7
+          doc: |
+            A list of EntryViews. Null means end of iteration.
```
ihsandemir, line 903: Does this mean that the client always needs to make the last fetch call which has to return null? Can we avoid this extra call? E.g. fi returned size is less than the requested batch size, it means, the entries are finished.

protocol-definitions/ReplicatedMap.yaml
```diff
@@ -852,3 +852,75 @@ methods:
...
+          type: List_ReplicatedMapEntryViewHolder
+          nullable: true
+          since: 2.7
+          doc: |
+            A list of EntryViews. Null means end of iteration. 
+  - id: 21
+    name: endEntryViewIteration
+    since: 2.7
+    doc: |
+      This method is used to release the resource generated by fetchEntryViews. Send this request after you retrieved the last page and no
```
ihsandemir, line 908: why do we need the client to make this request? Server knows it finished and it should release the resources. What is the rationale to make server cleanup depend on a client request?

ihsandemir (changes requested):

protocol-definitions/ReplicatedMap.yaml
frant-hartm: The memory per client iteration is constant.
If a request is retried with same cursor id it doesn't allocate new memory.

protocol-definitions/ReplicatedMap.yaml
frant-hartm: It's a UUID, it should be a non-issue.

protocol-definitions/ReplicatedMap.yaml
frant-hartm: This is the cursor id for the *next* batch.

protocol-definitions/ReplicatedMap.yaml
frant-hartm: We need this extra request to tell the server that the client has received the last page of data.

If we don't do this the following could happen:
- client sends a request - it is the last page
- server clears the cursor
- response with the last page gets lost
- client retries, but server doesn't have the data anymore.

However, I don't think we need a new request type for this, we can send another request when client receives batch < max batch size (or empty), the client knows the response to this will be empty, so it doesn't even need to wait for that 2nd response.

frant-hartm (changes requested):

protocol-definitions/ReplicatedMap.yaml
srknzl: You keep feeding the response cursor id to request cursor id to fetch the new pages. It's a simple id of the current page.

protocol-definitions/ReplicatedMap.yaml
srknzl: a UUID is unique so I don't need to say unique UUID. So a UUID is basically can be represented by a string, that's what I meant there. anyway

protocol-definitions/ReplicatedMap.yaml
srknzl: no you can't do that if the page size divides the total entry count evenly.

protocol-definitions/ReplicatedMap.yaml
srknzl: If page size divides entry count evenly, that logic won't work. I can add another parameter "endIteration:true" to the original method but I prefer a new method because it resembles the SQL close statement.

srknzl: I think by empty you mean null page. it can work but I don't see a problem in a new method

frant-hartm (approved):

ahmetmircik (approved):

srknzl:
@ahmetmircik @frant-hartm Apologies but I had to stale your reviews and change the design a bit after dealing with test failures. I've added the new "newIteration" request parameter that should be set to true if requesting creation of a new iteration. It can be safely retried if set to true, if there is a iteration with the cursor id, it won't create a new iteration. 

This mainly fixes the problem if a client sends a null cursor id (previous design) and does not receive the cursor id somehow. Then, it'll retry the request and create another completely different iterator. That's why the initial uuid is created on the client side now. 

I've also noticed considerable amount of bugs and flaky tests and fixed them all in the main pr.

srknzl:
Another thing I've ended up doing was one of your suggestions in the past. The response page is not nullable anymore and if received items in the page is less than the batchSize, the iteration ends. Note that client is responsible for clearing resources even if the iteration ends. A member only cleans it via the cleanup future

frant-hartm (approved):

ahmetmircik (approved):

ihsandemir (approved):
Approved based on other approvals.

PR #494, opened 2023-12-19
Add nullable validator

srknzl:
Related part in readme: 

* Note that for parameters of fixed size types (see `FixSizedTypes` in [currently supported types](binary/__init__.py))
`nullable` property is ignored. `UUID` is always nullable while other fixed size types (`boolean`, `byte`, `int`, `long`)
cannot be made nullable.

According to readme fixed size fields being nullable are ignored. I added it to validation code instead because we don't ignore it and generate "`@Nullable` int" etc that breaks the compat tests code.


util.py
```diff
@@ -471,6 +480,28 @@ def is_parameters_ordered_and_semantically_correct(since, name, params, protocol
...
+def contains_invalid_nullability_field(params):
+    for param in params:
+        # According to the readme, UUID fields should be nullable. But a lot of codecs violate that now,
+        # so skipping that check
+        """ 
+        if param["type"] == "UUID" and not param["nullable"]:
+            print(
+                'Check the nullable value of "%s" field of the "%s".\n'
+                "UUID fields should be nullable!"
+                % (param["name"], param["since"])
```
ihsandemir, line 492: Instead of `since`, you need the message name here?

srknzl: see https://github.com/hazelcast/hazelcast-client-protocol/pull/494/commits/80b844a7cd7e347f66a4b8d0b5fb0671f021ab64

ihsandemir (approved):

emreyigit (approved):

PR #495, opened 2023-12-21
Rename `namespace` to `userCodeNamespace` in various places

JamesHazelcast:
Updated as part of a wider effort to bring "UCD namespace" naming into consistency under "User Code Namespace".

Mono PR: https://github.com/hazelcast/hazelcast-mono/pull/272

### DIFF
--- a/protocol-definitions/DynamicConfig.yaml
+++ b/protocol-definitions/DynamicConfig.yaml
@@ -1430,7 +1430,7 @@ methods:
     name: removeUserCodeNamespaceConfig
     since: 2.7
     doc: |
-      Removes a namespace configuration.
+      Removes a user code namespace configuration.
     request:
       retryable: false
       partitionIdentifier: -1
--- a/protocol-definitions/DynamicConfig.yaml
+++ b/protocol-definitions/DynamicConfig.yaml
@@ -76,12 +76,12 @@ methods:
           since: 2.0
           doc: |
             Number of entries to be sent in a merge operation.
-        - name: namespace
+        - name: userCodeNamespace
           type: String
           nullable: true
           since: 2.7
           doc: |
-            Name of the namespace applied to this instance.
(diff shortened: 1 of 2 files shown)


protocol-definitions/DynamicConfig.yaml
```diff
@@ -1405,10 +1405,10 @@ methods:
             The WAN batch publisher configurations.
     response: {}
   - id: 19
-    name: addNamespaceConfig
+    name: addUserCodeNamespaceConfig
     since: 2.7
     doc: |
-      Adds a namespace configuration.
+      Adds a user code namespace configuration.
```
JackPGreen, line 1411: I think the doc on the parameters needs to change as well

protocol-definitions/DynamicConfig.yaml
```diff
@@ -1427,7 +1427,7 @@ methods:
             List of resource definitions.
     response: {}
   - id: 20
-    name: removeNamespaceConfig
+    name: removeUserCodeNamespaceConfig
```
JackPGreen, line 1430: I think the doc on the parameters needs to change as well

JackPGreen (changes requested):

protocol-definitions/DynamicConfig.yaml
JamesHazelcast: The doc is already updated, as shown in the diff above.

protocol-definitions/DynamicConfig.yaml
JamesHazelcast: Fixed this missing doc in https://github.com/hazelcast/hazelcast-client-protocol/pull/495/commits/5b1ae0e786997c22e4d1d837baa148dd03daef7e

JackPGreen (approved):

vbekiaris (approved):

PR #503, opened 2024-01-22
Add CPMap.putIfAbsent

gbarnett-hz:
Adds `putIfAbsent` for `CPMap` for parity with `IMap` so one can atomically associate a key with a value if the key does not exist.

Hazelcast Mono PR: https://github.com/hazelcast/hazelcast-mono/pull/495

### DIFF
--- a/protocol-definitions/CPMap.yaml
+++ b/protocol-definitions/CPMap.yaml
@@ -236,3 +236,44 @@ methods:
           since: 2.7
           doc: |
             True if key was associated with newValue, otherwise false.
+  - id: 7
+    name: putIfAbsent
+    since: 2.7
+    doc: |
+      Puts the key-value into the specified map if the key is not currently associated with a value.
+    request:
+      retryable: false
+      partitionIdentifier: -1
+      params:
+        - name: groupId
+          type: RaftGroupId
+          nullable: false
+          since: 2.7
+          doc: |
+            CP group ID of this CPMap instance.
+        - name: name
+          type: String
+          nullable: false
+          since: 2.7
+          doc: |
+            Name of this CPMap instance.
+        - name: key
+          type: Data
+          nullable: false
+          since: 2.7
+          doc: |
+            Key of the value.
+        - name: value
+          type: Data
+          nullable: false
+          since: 2.7
+          doc: |
+            Value to associate with the key.
+    response:
+      params:
+        - name: response
+          type: Data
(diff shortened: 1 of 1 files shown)


protocol-definitions/CPMap.yaml
```diff
@@ -236,3 +236,44 @@ methods:
...
           doc: |
             True if key was associated with newValue, otherwise false.
+  - id: 7
+    name: putIfAbsent
+    since: 2.7
+    doc: |
+      Puts the key-value into the specified map if the key is not currently associated with a value.
+    request:
+      retryable: false
+      partitionIdentifier: -1
```
ihsandemir, line 246: ? isn't it partition specific op?

protocol-definitions/CPMap.yaml
```diff
@@ -236,3 +236,44 @@ methods:
...
+          doc: |
+            Value to associate with the key.
+    response:
+      params:
+        - name: response
+          type: Data
+          nullable: true
+          since: 2.7
+          doc: |
+            Value associated with the key if already present.
```
ihsandemir, line 279:
```suggestion
            Value associated with the key if already present or null otherwise.
```

protocol-definitions/CPMap.yaml
gbarnett-hz: As discussed, all CP is same.

ihsandemir (approved):

protocol-definitions/CPMap.yaml
srknzl: hmm what does "all cp is same" mean?

protocol-definitions/CPMap.yaml
```diff
@@ -236,3 +236,44 @@ methods:
...
+          doc: |
+            Value to associate with the key.
+    response:
+      params:
+        - name: response
+          type: Data
+          nullable: true
+          since: 2.7
+          doc: |
+            Value associated with the key if already present, otherwise null.
```
srknzl, line 279: this is old value right? you may want to clarify it even more by adding old at the beginning

protocol-definitions/CPMap.yaml
gbarnett-hz: All CP ops use partition id of `-1`. Same with `CPMap` which is already there and in use.

srknzl (approved):

PR #507, opened 2024-02-05
Add consistency check strategy to WAN reference config [CN-1129]

SeriyBg:


protocol-definitions/MC.yaml
```diff
@@ -643,6 +643,14 @@ methods:
             0 - DISCARD_AFTER_MUTATION
             1 - THROW_EXCEPTION
             2 - THROW_EXCEPTION_ONLY_IF_REPLICATION_ACTIVE
+        - name: consistencyCheckStrategy
+          type: byte
```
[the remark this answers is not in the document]
SeriyBg: Well, I followed the Hazelcast config type, which is also `byte`. I assume we don't exclude the addition of a new consistency check strategy in the future, and with `bool` it won't be scalable

ihsandemir (approved):

protocol-definitions/MC.yaml
```diff
@@ -643,6 +643,14 @@ methods:
...
+          nullable: false
```
yuce, line 648: @ihsandemir  Can't remember the exact rules here, but wouldn't it be better to mark this as `nullable`?

ihsandemir: we do not mark primitives as nullable.

yuce (approved):
Looks good!

PR #508, opened 2024-02-05
Update `updateMapConfig` task with new parameter [HZ-4221]

JamesHazelcast:
We are expanding the `UpdateMapConfigOperation` to include the `WanReplicationRef` parameter.

Mono PR: https://github.com/hazelcast/hazelcast-mono/pull/544

### DIFF
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -224,19 +224,12 @@ methods:
             7 - USED_NATIVE_MEMORY_PERCENTAGE
             8 - FREE_NATIVE_MEMORY_SIZE
             9 - FREE_NATIVE_MEMORY_PERCENTAGE
-        - name: applyWanReplicationRef
-          type: boolean
-          nullable: false
-          since: 2.7
-          doc: |
-            Determines whether the supplied WanReplicationRef should be applied.
-            This is for backwards compatible versioning, as supplying a null WanReplicationRef is valid.
         - name: wanReplicationRef
           type: WanReplicationRef
           nullable: true
           since: 2.7
           doc: |
-            The new WanReplicationRef to apply, if applyWanReplicationRef is true.
+            The new WanReplicationRef to apply.
     response: {}
   - id: 5
     name: getMemberConfig
--- a/protocol-definitions/MC.yaml
+++ b/protocol-definitions/MC.yaml
@@ -224,6 +224,19 @@ methods:
             7 - USED_NATIVE_MEMORY_PERCENTAGE
             8 - FREE_NATIVE_MEMORY_SIZE
             9 - FREE_NATIVE_MEMORY_PERCENTAGE
+        - name: applyWanReplicationRef
+          type: boolean
+          nullable: false
+          since: 2.7
+          doc: |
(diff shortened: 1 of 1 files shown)


vbekiaris (approved):
The changes look good to me

protocol-definitions/MC.yaml
```diff
@@ -224,6 +224,19 @@ methods:
             7 - USED_NATIVE_MEMORY_PERCENTAGE
             8 - FREE_NATIVE_MEMORY_SIZE
             9 - FREE_NATIVE_MEMORY_PERCENTAGE
+        - name: applyWanReplicationRef
+          type: boolean
+          nullable: false
```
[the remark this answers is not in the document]
ihsandemir: Yes, definitely you can set `nullable` for any new field. For primitives, we always set it to false by the way.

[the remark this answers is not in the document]
JamesHazelcast: > We always differentiate existence of a new field using another generated boolean flag is<field name>Exist.

This makes sense, my codecs generated these fields - so to clarify, this PR is fine to merge as-is then @ihsandemir?

protocol-definitions/MC.yaml
```diff
@@ -224,6 +224,19 @@ methods:
...
+          since: 2.7
+          doc: |
+            Determines whether the supplied WanReplicationRef should be applied.
+            This is for backwards compatible versioning, as supplying a null WanReplicationRef is valid.
+        - name: wanReplicationRef
+          type: WanReplicationRef
+          nullable: true
```
ihsandemir, line 236: since this is nullable, do you still need the `applyWanReplicationRef` field?

JamesHazelcast: Yes, because a null `WanReplicationRef` is valid (removes the existing reference from the `MapConfig`) - so to determine if we have an actual `null` value (meaning remove the reference), or if it's just behind on protocol, the `applyWanReplicationRef` value is used (`false` in older protocols, `true` in newer). I plan to remove this value in future releases as it's just for backwards compatibility.

ihsandemir: So, the wanReplicationRef field is also new, and you can check existence of this new field. if the field exists(i do not mean null) then it is from new version client and you apply it. Otherwise field does not exist and it is old version client. Hence, i still could not make sense of the other field. May be i miss some context.

JamesHazelcast: I think I see what you're suggesting - so you think that [the `UpdateMapConfigOperation`](https://github.com/hazelcast/hazelcast-mono/pull/544/files#diff-60b54afbd749e10f76f9c7595e9a462d20e90f0dd73a9c8e0b99c3203fb71067) should still continue to use a `applyWanReplicationRef` field that it controls locally (if sending version is <5.4, it sets to `false`, otherwise `true`), but we don't need to send it over the wire/add to codec?

I think that could work, but please let me know if you're expecting differently.

ihsandemir (approved):

vbekiaris (approved):

ahmetmircik (approved):

PR #509, opened 2024-02-12
Add the cluster meta data to authentication response [API-2171]

ihsandemir:
Added the cluster meta data, the member list and partition table information at the client authentication response so that the client does not have to wait the next cluster view event to start or continue the operation.

hazelcast pr: https://github.com/hazelcast/hazelcast-mono/pull/772

### DIFF
--- a/protocol-definitions/Client.yaml
+++ b/protocol-definitions/Client.yaml
@@ -278,25 +278,25 @@ methods:
         - name: memberListVersion
           type: int
           nullable: false
-          since: 2.0
+          since: 2.7
           doc: |
             Incremental member list version
         - name: memberInfos
           type: List_MemberInfo
           nullable: false
-          since: 2.0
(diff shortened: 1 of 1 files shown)


protocol-definitions/Client.yaml
```diff
@@ -130,6 +130,30 @@ methods:
...
+        - name: memberListVersion
+          type: int
+          nullable: false
+          since: 2.0
+          doc: |
+            Incremental member list version
+        - name: memberInfos
+          type: List_MemberInfo
+          nullable: false
+          since: 2.0
```
srknzl, line 142: 2.7

ihsandemir: fixed.

emreyigit (approved):

emreyigit (approved):

PR #511, opened 2024-02-21
Implement ClientClusterService.getClusterVersion() [HZ-4446]

burakgok:
Mono PR: hazelcast/hazelcast-mono#871


srknzl (commented):
cpp, cs and ts language generations fail. can you fix them?

burakgok:
verify

burakgok:
verify

burakgok:
verify

protocol-definitions/custom/Custom.yaml
```diff
@@ -1499,3 +1499,15 @@ customTypes:
         type: String
         since: 2.7
         nullable: true
+  - name: Version
+    since: 2.7
+    returnWithFactory: true
+    params:
```
ihsandemir, line 1505: Very similar to `MemberVersion` (assume patch field is 0). Do we really need yet another new custom type?

protocol-definitions/custom/Custom.yaml
```diff
@@ -1499,3 +1499,15 @@ customTypes:
         type: String
         since: 2.7
         nullable: true
+  - name: Version
```
ihsandemir, line 1502:
```suggestion
  - name: ClusterVersion
```

protocol-definitions/custom/Custom.yaml
burakgok: The cluster version is not `MemberVersion`, and I am not introducing another type, just adding an existing type to the client protocol. The WAN version is also a `Version`, i.e. it consists of major and minor versions only, so `Version` may have other usages in the _near_ future, considering that we try to send more cluster information to the clients.

protocol-definitions/custom/Custom.yaml
burakgok: https://github.com/hazelcast/hazelcast-client-protocol/pull/511#discussion_r1519414404

srknzl (approved):

JamesHazelcast (approved):

protocol-definitions/custom/Custom.yaml
```diff
@@ -1499,3 +1499,15 @@ customTypes:
...
+    since: 2.7
```
ihsandemir, line 1503:
```suggestion
    since: 2.8
```
For 5.5 release, we need to increment the version.

burakgok: 74593a07

PR #512, opened 2024-03-21
Vector collection config for java

shultseva:
codes for VectorCollectionConfig and VectorIndexConfig

### DIFF
--- a/protocol-definitions/DynamicConfig.yaml
+++ b/protocol-definitions/DynamicConfig.yaml
@@ -1428,7 +1428,7 @@ methods:
     response: {}
   - id: 20
     name: addVectorCollectionConfig
-    since: 2.0
+    since: 2.8
     doc: |
       Adds a new vector collection configuration to a running cluster.
     request:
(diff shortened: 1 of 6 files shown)


protocol-definitions/DynamicConfig.yaml
```diff
@@ -1426,3 +1426,25 @@ methods:
           doc: |
             List of resource definitions.
     response: {}
+  - id: 20
+    name: addVectorCollectionConfig
+    since: 2.0
```
k-jamroz, line 1431: `since: 2.8`?

k-jamroz (approved):

k-jamroz:
Regarding the pr-builder failure - I think we should not generate vector classes in typescript. I don't know if there is an easy way to disable it without hacking `util.py`. @yuce WDYT? Maybe we should ignore such errors in similar way as in https://github.com/hazelcast/hazelcast-client-protocol-internal/pull/1. I am not sure if that is a good long-term solution though.

k-jamroz:
> Regarding the pr-builder failure - I think we should not generate vector classes in typescript.

This actually is pretty easy: https://github.com/hazelcast/hazelcast-client-protocol-internal/pull/1/commits/2aa496cadb892c74730c66db9d7314575ab8f35f

vbekiaris (approved):

yuce (approved):

PR #516, opened 2024-04-12
Additional params for index vector [HZAI-57]

shultseva:
added params:
```
private int maxDegree;
private int beamWidth\efConstruction;
private boolean useDeduplication;
```

### DIFF
--- a/protocol-definitions/custom/Custom.yaml
+++ b/protocol-definitions/custom/Custom.yaml
@@ -1518,15 +1518,15 @@ customTypes:
       - name: maxDegree
         type: int
         since: 2.8
-        nullable: true
+        nullable: false
       - name: efConstruction
         type: int
         since: 2.8
-        nullable: true
+        nullable: false
       - name: useDeduplication
         type: boolean
         since: 2.8
-        nullable: true
+        nullable: false
 
   - name: VectorPair
     since: 2.8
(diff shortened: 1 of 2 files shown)


protocol-definitions/custom/Custom.yaml
```diff
@@ -1515,6 +1515,18 @@ customTypes:
         type: int
         since: 2.8
         nullable: false
+      - name: maxDegree
+        type: int
+        since: 2.8
+        nullable: true
```
k-jamroz, line 1521: `FixSizedTypes` fields cannot be nullable (see readme - I was surprised by that)

shultseva: The nullable int value is true or false doesn't impact the generated codec. Additionally, nullable int is already utilized in this file.  So I prefer to leave it as this to maintain alignment with the mono-repo documentation.

k-jamroz: `nullable: true` is ignored here (only for fixed sized types!) and works like `nullable: false`. Do not introduce misleading, ignored setting. If it is used elsewhere for such types, then it is not good either, but with quick search I did not find any such case. Do you have an example of that?

shultseva: 0c1f574e56aea824fd1f55bb529f2f6ea4ddebec

k-jamroz (approved):

vbekiaris (approved):

PR #518, opened 2024-04-23
Cleanup index operation

shultseva:
Cleanup index operation

### DIFF
--- a/protocol-definitions/VectorCollection.yaml
+++ b/protocol-definitions/VectorCollection.yaml
@@ -242,3 +242,26 @@ methods:
           doc: |
             Zero or more VectorSearchResult values.
 
+  - id: 9
+    name: cleanup
+    since: 2.8
+    doc: |
+      Cleanup index.
+    request:
+      retryable: true
+      partitionIdentifier: none
+      params:
+        - name: name
+          type: String
+          nullable: false
+          since: 2.8
+          doc: |
+            Name of the Vector Collection.
+        - name: index
+          type: String
+          nullable: true
+          since: 2.8
+          doc: |
+            Name of the Index to cleanup.
+    response: {}
+


protocol-definitions/VectorCollection.yaml
```diff
@@ -242,3 +242,26 @@ methods:
...
+      retryable: true
+      partitionIdentifier: none
+      params:
+        - name: name
+          type: String
+          nullable: false
+          since: 2.8
+          doc: |
+            Name of the Vector Collection.
+        - name: index
```
k-jamroz, line 260: `indexName`?

protocol-definitions/VectorCollection.yaml
```diff
@@ -242,3 +242,26 @@ methods:
           doc: |
             Zero or more VectorSearchResult values.
 
+  - id: 9
+    name: cleanup
```
k-jamroz, line 246: 1. how should we name the operation? `cleanup` is JVector specific and IMO too similar to `clear`. `vaccum` is interesting, but may not be understood. `optimize`?
2. should the operation be per index or per collection? IMO per index (as you implemented) is as it is more granular. However, are there many cases with multiple indexes when only one index requires cleanup? Maybe others have different opinion?

k-jamroz: BTW, we could address point 2 by defining that null index name means _all_ indexes instead of _the only_ index.

vbekiaris: `vacuum` is probably only understood by persons with specific DB background... `optimize` seems best option.
I think it makes sense that `null` implies all indexes instead of the only index. If I put thousands of new entries and I want to perform cleanup, I need to do it on all indexes (since all have new entries), not just one.

shultseva: In my view, it's exceedingly uncommon to require a complete cleanup of all indexes. 
Furthermore, processing such a task is exceptionally long, and, if necessary, it's advisable to handle each cleanup individually and under control.

shultseva: Moreover, it doesn't align with other methods; searching without an index implies searching in the first and only index, and putting without an index means putting in the first and only vector. 
However, cleaning up without specifying an index implies cleaning up all indexes. 

My suggestion is to leave it as is. If at some point we determine that we require methods to clean up all indexes, then we can consider creating an additional method specifically for that purpose, like a "cleanupAll" method.

protocol-definitions/VectorCollection.yaml
```diff
@@ -242,3 +242,27 @@ methods:
...
+          since: 2.8
+          doc: |
+            Name of the Vector Collection.
+        - name: indexName
+          type: String
+          nullable: true
+          since: 2.8
+          doc: |
+            Name of the Index to optimize.
+            A null value signifies the optimization of the first and only index within the collection.
```
k-jamroz, line 266:
```suggestion
            A null value triggers the optimization of the only index within the collection.
```
if it is only, it is first

vbekiaris (approved):

k-jamroz (approved):

PR #524, opened 2024-06-04
Clear method api [HZAI-32]

shultseva:
Clear vector collection method api

### DIFF
--- a/protocol-definitions/VectorCollection.yaml
+++ b/protocol-definitions/VectorCollection.yaml
@@ -272,3 +272,19 @@ methods:
             A null value triggers the optimization of the only index within the collection.
     response: {}
 
+  - id: 10
+    name: clear
+    since: 2.8
+    doc: |
+      Clear vector collection.
+    request:
+      retryable: true
+      partitionIdentifier: none
+      params:
+        - name: name
+          type: String
+          nullable: false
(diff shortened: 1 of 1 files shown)


k-jamroz (approved):

protocol-definitions/VectorCollection.yaml
```diff
@@ -272,3 +272,19 @@ methods:
             A null value triggers the optimization of the only index within the collection.
     response: {}
 
+  - id: 10
+    name: clear
+    since: 2.8
+    doc: |
+      Clear vector collection.
+    request:
+      retryable: true
```
k-jamroz, line 281: I wonder if it should be retryable. IMap.clear is not

vbekiaris: As a mutating operation, it's safer to go with `false`

vbekiaris (changes requested):
@shultseva I think `retryable` value should be `false`

protocol-definitions/VectorCollection.yaml
shultseva: 03064e0df406a2e64f6b378874f28fece5a9da18

vbekiaris (approved):

PR #525, opened 2024-06-12
Add client protocol for member demotion [MC-2855]

emrberk:


promanenko (approved):

userMuratt (approved):

protocol-definitions/MC.yaml
```diff
@@ -955,3 +955,12 @@ methods:
           doc: |
             true if the session is found & closed,
             false otherwise.
+  - id: 37
+    name: demoteDataMember
+    since: 2.8
+    doc: |
+      Demotes the data member it's called on to a lite member.
+    request:
+      retryable: false
```
ihsandemir, line 964: why not retryable? is it not idempotent?

protocol-definitions/MC.yaml
```diff
@@ -955,3 +955,12 @@ methods:
           doc: |
             true if the session is found & closed,
             false otherwise.
+  - id: 37
+    name: demoteDataMember
```
ihsandemir, line 959: can we have a more descriptive name if it is demoting to a lite member, e.g.  `changeToLiteMember` or `demoteToLiteMember` ?

protocol-definitions/MC.yaml
```diff
@@ -955,3 +955,12 @@ methods:
...
+    since: 2.8
+    doc: |
+      Demotes the data member it's called on to a lite member.
+    request:
+      retryable: false
+      partitionIdentifier: -1
+    response: {}
```
ihsandemir, line 966: how do we communicate success and failure? If communicate failure with exceptions, can you document what are possible exceptions as well?

protocol-definitions/MC.yaml
emrberk: Should we retry demoting the member after exception is thrown? I thought it won't add up any value.

protocol-definitions/MC.yaml
promanenko: it's consistent with the existing `promoteLiteMember`

protocol-definitions/MC.yaml
promanenko: we should not retry it

protocol-definitions/MC.yaml
emrberk: Exception is thrown in case of a failure. Added to the doc part.

zhmaeff (approved):

PR #545, opened 2025-04-11
Add autoStepDownWhenLeader to MemberInfo [CORE-216]

gareth-johnston:
Adds `autoStepDownWhenLeader` boolean added to MemberInfo

### DIFF
--- a/protocol-definitions/custom/Custom.yaml
+++ b/protocol-definitions/custom/Custom.yaml
@@ -828,10 +828,6 @@ customTypes:
         type: Map_EndpointQualifier_Address
         nullable: false
         since: 2.0.1
-      - name: autoStepDownWhenLeader
-        type: boolean
-        nullable: false
-        since: 2.9
   - name: EndpointQualifier
     returnWithFactory: true
     since: 2.0.1
--- a/protocol-definitions/custom/Custom.yaml
+++ b/protocol-definitions/custom/Custom.yaml
@@ -831,7 +831,7 @@ customTypes:
       - name: autoStepDownWhenLeader
         type: boolean
         nullable: false
-        since: 2.8.1
+        since: 2.9
   - name: EndpointQualifier
     returnWithFactory: true
     since: 2.0.1
@@ -989,6 +989,10 @@ customTypes:
         type: Address
         nullable: false
         since: 2.1
+      - name: autoStepsDownWhenLeader
+        type: boolean
+        nullable: false
+        since: 2.9
   - name: MigrationState
     since: 2.2
     params:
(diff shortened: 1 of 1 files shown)


protocol-definitions/custom/Custom.yaml
```diff
@@ -828,6 +828,10 @@ customTypes:
         type: Map_EndpointQualifier_Address
         nullable: false
         since: 2.0.1
+      - name: autoStepDownWhenLeader
```
ihsandemir, line 831:
```suggestion
      - name: cpAutoStepDownWhenLeader
```

protocol-definitions/custom/Custom.yaml
```diff
@@ -985,6 +989,10 @@ customTypes:
         type: Address
         nullable: false
         since: 2.1
+      - name: autoStepDownWhenLeader
```
ihsandemir, line 992: why duplicated the same info in the CPMember? Can you not get the member info using the uuid and check this field in MemberInfo?

ihsandemir (dismissed):

protocol-definitions/custom/Custom.yaml
ahmetmircik: agree that prefixing with `cp` makes it more understandable, especially for the future readers.

ahmetmircik (approved):

JamesHazelcast (approved):
I agree with @ihsandemir that prefixing with `cp` makes sense, to provide extra clarity - approving in advance though 👍

gareth-johnston:
> I agree with @ihsandemir that prefixing with `cp` makes sense, to provide extra clarity - approving in advance though 👍

Do you think that is still relevant now? That comment was relating to when we were changing `memberInfo` -  absolutely in that case. But now we are only changing CPMember -- I believe it is clear enough in this case, do you feel strongly?

JamesHazelcast:
> Do you think that is still relevant now? That comment was relating to when we were changing `memberInfo` - absolutely in that case. But now we are only changing CPMember -- I believe it is clear enough in this case, do you feel strongly?

Ah, apologies - no I don't think it's relevant in that case, all good to go from my POV

yuce (approved):

protocol-definitions/custom/Custom.yaml
```diff
@@ -985,6 +985,10 @@ customTypes:
...
+        type: boolean
+        nullable: false
+        since: 2.9
```
ihsandemir, line 991: This should be 2.10 for 5.7. 2.9 was used for platform 5.6.

PR #546, opened 2025-04-15
Move Dynamic `setDiagnosticsConfig` to MC space [DEX-323]

emreyigit:
`setDiagnosticsConfig` dynamic config API is moved to MC codec space. Since client protocol 2.9 has not been released yet, there is no backward compatibility issue.


gareth-johnston (approved):

protocol-definitions/MC.yaml
```diff
@@ -981,3 +981,61 @@ methods:
...
+          nullable: false
+          since: 2.9
+          doc: |
+            The maximum number of rolling files to keep on disk
+        - name: logDirectory
+          type: String
+          nullable: false
+          since: 2.9
+          doc: |
+            The output directory of the diagnostics log files
```
ihsandemir, line 1028: Can you indicate if this is a file path or not and if this is relative path or absolute path? Can you also may be put an example of such a path.

protocol-definitions/MC.yaml
```diff
@@ -981,3 +981,61 @@ methods:
...
+          nullable: true
+          since: 2.9
+          doc: |
+            The prefix for the diagnostics file
+        - name: properties
+          type: Map_String_String
+          nullable: true
+          since: 2.9
+          doc: |
+            Properties for the diagnostic plugins
```
ihsandemir, line 1040: The properties includes entries specific to each individual plugin. e.g. .... for ... plugin

protocol-definitions/MC.yaml
```diff
@@ -981,3 +981,61 @@ methods:
...
+    response: {}
```
ihsandemir, line 1041: Can you comment on the response? E.g. This message changes diagnostics cluster wide and if one member fails, what happens? Which exceptions do we expect?

protocol-definitions/MC.yaml
emreyigit: fixed

protocol-definitions/MC.yaml
emreyigit: fixed.

protocol-definitions/MC.yaml
emreyigit: fixed.

protocol-definitions/MC.yaml
ihsandemir: can you also write if it is a relative or absolute path or may be both?

emreyigit: fixed

ihsandemir (approved):

PR #547, opened 2025-05-02
Mark mapName field for QueryCacheEventData as nullable [HZG-385]

JamesHazelcast:
Mono-repo PR: https://github.com/hazelcast/hazelcast-mono/pull/4286

### DIFF
--- a/protocol-definitions/custom/Custom.yaml
+++ b/protocol-definitions/custom/Custom.yaml
@@ -444,7 +444,7 @@ customTypes:
         since: 2.0
       - name: mapName
         type: String
-        nullable: false
+        nullable: true
         since: 2.7
   - name: QueueStoreConfigHolder
     since: 2.0


TomaszGaweda (approved):

protocol-definitions/custom/Custom.yaml
```diff
@@ -444,7 +444,7 @@ customTypes:
         since: 2.0
       - name: mapName
         type: String
-        nullable: false
+        nullable: true
```
ihsandemir, line 447: This will break compatibility, right? you change binary signature and old clients will fail.

ihsandemir (changes requested):

protocol-definitions/custom/Custom.yaml
ihsandemir: in which cases do we send null mapName?

JamesHazelcast: You're right, I had forgotten about older clients which will not expect `null` values - in that case, back to the drawing board for this one 😅

PR #549, opened 2025-05-14
Add the logger parameter to event handler constructions for C++ codecs [HZG-188]

ihsandemir:
Added the logger parameter to event handler constructions.

Also added better diagnostics for NotImplementedError messages.

### DIFF
--- a/cpp/__init__.py
+++ b/cpp/__init__.py
@@ -24,7 +24,7 @@ def cpp_types_encode(key):
     except KeyError:
         cpp_type = _cpp_types_common[key]
     if cpp_type == "NA":
-        raise NotImplementedError("Missing type Mapping")
+        raise NotImplementedError("Missing type Mapping for " + key)
     return cpp_type
 
 
@@ -34,7 +34,7 @@ def cpp_types_decode(key):
     except KeyError:
         cpp_type = _cpp_types_common[key]
     if cpp_type == "NA":
-        raise NotImplementedError("Missing type Mapping")
+        raise NotImplementedError("Missing type Mapping for " + key)
     return cpp_type
 
 
@@ -42,7 +42,7 @@ def get_size(type):
     try:
         size = _type_size[type]
     except KeyError:
-        raise NotImplementedError("Missing type size mapping")
+        raise NotImplementedError("Missing type size mapping " + type)
     return size
 
 
--- a/cpp/__init__.py
+++ b/cpp/__init__.py
@@ -42,7 +42,7 @@ def get_size(type):
     try:
         size = _type_size[type]
     except KeyError:
-        raise NotImplementedError("Missing type size mapping " + type)
+        raise NotImplementedError("Missing type size Mapping " + type)
     return size
 
 
--- a/cpp/codec-template.cpp.j2
+++ b/cpp/codec-template.cpp.j2
@@ -48,12 +48,15 @@
 
                 {# EVENTS#}
                 {% if method.events|length != 0 %}
+                {{ service_name.lower() }}_{{ method.name.lower() }}_handler::{{ service_name.lower() }}_{{ method.name.lower() }}_handler(logger& l)
+                    : impl::BaseEventHandler(l) {}
+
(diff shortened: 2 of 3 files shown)


cpp/__init__.py
```diff
@@ -34,15 +34,15 @@ def cpp_types_decode(key):
...
+        raise NotImplementedError("Missing type Mapping for " + key)
     return cpp_type
 
 
 def get_size(type):
     try:
         size = _type_size[type]
     except KeyError:
-        raise NotImplementedError("Missing type size mapping")
+        raise NotImplementedError("Missing type size mapping " + type)
```
JackPGreen, line 45:
```suggestion
        raise NotImplementedError("Missing type size Mapping " + type)
```

turbo nit - consistency with the other methods' exceptions.

cpp/codec-template.cpp.j2
```diff
@@ -48,12 +48,15 @@
...
+                {{ service_name.lower() }}_{{ method.name.lower() }}_handler::{{ service_name.lower() }}_{{ method.name.lower() }}_handler(logger& l)
+                    : impl::BaseEventHandler(l) {}
+
                 void {{ service_name.lower() }}_{{ method.name.lower() }}_handler::handle(ClientMessage &msg) {
                     auto messageType = msg.get_message_type();
                 {% for event in method.events%}
                     if (messageType == {{ event.id }}) {
                         {% if fixed_params(event.params)|length > 0 %}
-                        auto *initial_frame = reinterpret_cast<ClientMessage::frame_header_t *>({% endif %}{% if fixed_params(event.params)|length == 0 %}                        {% endif %}msg.rd_ptr(ClientMessage::EVENT_HEADER_LEN){% if fixed_params(event.params)|length > 0 %}){% endif %};
+                        auto *initial_frame = reinterpret_cast<ClientMessage::frame_header_type *>({% endif %}{% if fixed_params(event.params)|length == 0 %}                        {% endif %}msg.rd_ptr(ClientMessage::EVENT_HEADER_LEN){% if fixed_params(event.params)|length > 0 %}){% endif %};
```
JackPGreen, line 59: For my own sanity - on the client side, `frame_header_t` was renamed to `frame_header_type` in https://github.com/hazelcast/hazelcast-cpp-client/pull/995 (more specifically, [here](https://github.com/hazelcast/hazelcast-cpp-client/pull/995#discussion_r976644782)) but this change wasn't implemented upstream in the protocol?

So not _directly_ related to this change.

cpp/codec-template.cpp.j2
```diff
@@ -48,12 +48,15 @@
 
                 {# EVENTS#}
                 {% if method.events|length != 0 %}
+                {{ service_name.lower() }}_{{ method.name.lower() }}_handler::{{ service_name.lower() }}_{{ method.name.lower() }}_handler(logger& l)
+                    : impl::BaseEventHandler(l) {}
```
JackPGreen, line 52: The actual change in this PR is to add a new constructor for these `_handlers` that takes in a `logger` to ensure one is available - and because of this, we can access via a reference, not a pointer.

<sup>_(this is just me making sure I understand!)_<sup>

JackPGreen (approved):

cpp/codec-template.cpp.j2
ihsandemir: need to track the PRs to find the whole story but it is most possible that the codecs are not either fully generated since that change.

cpp/codec-template.cpp.j2
ihsandemir: yes, that is correct.

emreyigit (approved):
Approving but it's not trivial to grasp everything from here. I think you should also regenerate the all codecs for a better alignment as mentioned.

ihsandemir:
> Approving but it's not trivial to grasp everything from here. I think you should also regenerate the all codecs for a better alignment as mentioned.

yes, i will do it in a separate PR.

PR #551, opened 2025-05-27
C++ template: Changed the log level to finest when an unknown type new event arrives DEX-439

ihsandemir:
Changed the log level to finest when a new event arrives whose type the client does not know. It can be a newly added event to existing events. This is in sync with [Java client](https://github.com/hazelcast/hazelcast-client-protocol/blob/master/java/codec-template.java.j2#L402).

Also, updated the remove list and some type mappings.


cpp/__init__.py
JackPGreen, line 1: I don't understand the changes in this file. Could you explain, please?

emreyigit (approved):

JackPGreen:
@ihsandemir re-opened, I think we need this? Or at least, this would be beneficial for future stability?

JackPGreen (approved):

PR #568, opened 2026-01-26
Update docs generation instructions

Rob-Hazelcast:


md/__init__.py
JackPGreen, line 1: Bit confused - does it even make a PDF output?

Rob-Hazelcast: The main code generator tool (described in the README) generates markdown. These instructions are to convert the markdown file to a PDF. It works. But it's not obvious we need to provide these instructions, and IDK why they're in this particular file.

JackPGreen (approved):

jgardiner68 (approved):

PR #570, opened 2026-08-13
More Node.js client updates

yuce:


dmitriyrazboev (approved):

util.py
```diff
@@ -730,7 +730,8 @@ def create_environment(lang, namespace):
     env.globals["item_type"] = item_type
     env.globals["key_type"] = key_type
     env.globals["value_type"] = value_type
-    env.globals["namespace"] = namespace
+    if lang in [SupportedLanguages.JAVA, SupportedLanguages.CS]:
+        env.globals["namespace"] = namespace
```
ihsandemir, line 734: nit: does not seem to be node.js related.

ihsandemir (approved):

util.py
yuce: The `namespace` is a special Jinja variable. I have to use it to store some state during template rendering. Setting it in `env.globals` prevents me using it.

